PR #4, opened 2024-09-23
update husky config

groninge01:


apps/frontend-v3/package.json
garethfuller, line 1: These changes aren't required?

package.json
garethfuller, line 1: Same with these?

garethfuller (changes requested):

apps/frontend-v3/package.json
groninge01: No, they are not required. I just like a package.json to be sorted :)
As pnpm doesn't really do this maybe we should add a prettier plugin for this?

groninge01: Sorted the package.json.

package.json
groninge01: Sorted the package.json.

garethfuller (approved):

PR #5, opened 2024-09-24
chore: use Node.js subpath imports

groninge01:


groninge01:
this needs more work, closing for now

PR #10, opened 2024-10-15
feat: add beets app

groninge01:


package.json
```diff
@@ -2,9 +2,9 @@
   "name": "@balancer/frontend-monorepo",
   "private": true,
   "scripts": {
-    "build": "turbo build",
-    "build:dev": "turbo build:dev",
-    "dev": "turbo dev",
+    "build": "turbo build --filter frontend-v3",
+    "build:dev": "turbo build:dev --filter frontend-v3",
+    "dev": "turbo dev --filter frontend-v3",
```
groninge01, line 7: as this will only be used locally i think we can filter here (or just the 'dev' task)

apps/beets-frontend-v3/.husky/pre-commit
garethfuller, line 1: Should husky not be hoisted up to top level? Same for frontend-v3

apps/beets-frontend-v3/.stylelintignore
garethfuller, line 1: Same with stylelint, should be top level config.

package.json
garethfuller: Like discussed, please don't do this for now. Feel free to add dev:bal and dev:beets to turbo.json to run specific apps locally.

apps/beets-frontend-v3/.husky/pre-commit
groninge01: Moved husky to root and also added a custom lint-staged config for the lib package.

package.json
groninge01: Updated this.

apps/beets-frontend-v3/.stylelintignore
groninge01: removed stylelint configs from app folders

garethfuller (changes requested):
Please update readme to include the beets app in the list of apps and explain that there are separate apps for spinning up beets vs bal.

apps/beets-frontend-v3/.prettierrc.cjs
garethfuller, line 1: Should prettier config not also be at the root level?

garethfuller (approved):

PR #65, opened 2024-10-15
fix: show 0% when there are no surplus apr items

### LINKED ISSUE #62: Show 0% LVR
When LVR is 0% show it in APR tooltip.

Image

groninge01:
fixes #62 

when there are no surplus apr items show LVR as 0% for cow amm pools

image

### DIFF
--- a/packages/lib/shared/components/tooltips/apr-tooltip/BaseAprTooltip.tsx
+++ b/packages/lib/shared/components/tooltips/apr-tooltip/BaseAprTooltip.tsx
@@ -185,15 +185,13 @@ function BaseAprTooltip({
           tooltipText={merklIncentivesTooltipText}
         />
       )}
-      {hasSurplusIncentives && (
-        <TooltipAprItem
-          {...basePopoverAprItemProps}
-          displayValueFormatter={usedDisplayValueFormatter}
-          title="Prevented LVR"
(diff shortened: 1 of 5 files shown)


packages/lib/shared/components/tooltips/apr-tooltip/BaseAprTooltip.tsx
```diff
@@ -185,12 +187,12 @@ function BaseAprTooltip({
...
         />
       )}
-      {hasSurplusIncentives && (
+      {isCowAmmPool(poolType) && (
         <TooltipAprItem
           {...basePopoverAprItemProps}
           displayValueFormatter={usedDisplayValueFormatter}
           title="Prevented LVR"
-          apr={surplusIncentivesAprDisplayed}
+          apr={hasSurplusIncentives ? surplusIncentivesAprDisplayed : bn(0)}
```
garethfuller, line 195: Do we need `hasSurplusIncentives` here? I think `surplusIncentivesAprDisplayed` will just be zero right?

groninge01: You're right. Removed it.

garethfuller (approved):

PR #69, opened 2024-10-16
chore: duplicated sdk

agualis:
b-sdk was declared twice. Keep the lib dependency and remove the frontend-v3 one.

TODO: remove other duplicated dependencies.

### DIFF
--- a/apps/frontend-v3/package.json
+++ b/apps/frontend-v3/package.json
@@ -19,7 +19,6 @@
   },
   "dependencies": {
     "@apollo/client": "^3.11.8",
-    "@balancer/sdk": "^0.28.1",
     "@chakra-ui/anatomy": "^2.2.2",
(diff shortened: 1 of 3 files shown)


agualis:
Closing because it seems to be that best practice in turborepo is to install in each package/app where it's needed: https://turbo.build/repo/docs/crafting-your-repository/managing-dependencies

> Enhanced flexibility: In a monorepo at scale, it can be unrealistic to expect each package to use the same version of an external dependency. When there are many teams working in the same codebase, there will be differing priorities, timelines, and needs due to the realities of operating at scale. By installing dependencies in the package that uses them, you can enable your ui team to bump to the latest version of TypeScript, while your web team can prioritize shipping new features and bumping TypeScript later. Additionally, if you still want to keep dependency versions in sync, you can do that, too.

PR #72, opened 2024-10-16
feat: add prettier-config package

groninge01:
- updated the eslint config according to [this](https://github.com/prettier/eslint-config-prettier)
- reworked lintstaged configs (maybe it also needs a package as there is some duplication now)


.eslintignore
```diff
@@ -1,4 +1,7 @@
 node_modules/
 .next/
 **/generated/*.ts
-theme.ts
\ No newline at end of file
+theme.ts
+packages/prettier-config/
+packages/eslint-config/
+packages/typescript-config/
```
groninge01, line 7: as these files won't be updated much i think we can ignore then for now?
otherwise you'd need to add the eslint & prettier configs to each

garethfuller: Can they not cover themselves?

groninge01: my thoughts:
- these files are pretty static so i wonder if justifies the overhead
- would we run into circular dependencies if we do add a package as it's own dependency?

i will look into it

groninge01: customized eslint & lint-staged to also cover the config packages

packages/lib/modules/pool/PoolList/PoolListFilters.tsx
```diff
@@ -64,7 +64,7 @@ export function useFilterTagsVisible() {
 
   useEffect(() => {
     setIsVisible(
-      networks.length > 0 || poolTypes.length > 0 || minTvl > 0 || poolCategories.length > 0
+      networks.length > 0 || poolTypes.length > 0 || minTvl > 0 || poolCategories.length > 0,
```
garethfuller, line 67: This seems like an incorrect change?

garethfuller (changes requested):

packages/lib/modules/pool/PoolList/PoolListFilters.tsx
groninge01: the default was changed in prettier v3: https://prettier.io/docs/en/options.html#trailing-commas

garethfuller: Yeah that's fine but this particular usage looks weird to me. `aFunction(someParam,)` does not look right?

garethfuller: Whatever part of the rule is causeing that, I think we should change. Trailing commas for arrays and objects are fine.

groninge01: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas#trailing_commas_in_functions

garethfuller: I'm not a fan, but if you want to keep it `all` instead of `es5` then let's do it. We can easily change later.

garethfuller (approved):

packages/lib/modules/pool/PoolList/PoolListFilters.tsx
groninge01: added `trailingComma: "es5",` to prettier config

PR #73, opened 2024-10-16
fix: use fresh swapState when updaring swapStateVar

agualis:
We had this bug: 
1. Go to swaps page and select `swap/gnosis/USDC/COW`
2. Manually update the url to `swap/gnosis/xDAI/COW` and hit ENTER
3. Bug: the tokenIn remains USDC when it should be xDAI

The reason is that the functions that udpate `swapStateVar` were using `swapState` without making sure that it had the most recent updated value, causing different issues when performing consecutive updates like in:
``` ts
    setInitialChain(chain)
    setInitialTokenIn(tokenIn)
    setInitialTokenOut(tokenOut)
    setInitialAmounts(amountIn, amountOut)
```

The solution is enforcing that `swapState` has the most last updated value with  `const swapState = swapStateVar()` before updating it again.

### DIFF
--- a/packages/lib/modules/swap/SwapProvider.tsx
+++ b/packages/lib/modules/swap/SwapProvider.tsx
@@ -174,6 +174,7 @@ export function _useSwap({ urlTxHash, ...pathParams }: PathParams) {
   })
 
   function handleSimulationResponse({ returnAmount, swapType }: SimulateSwapResponse) {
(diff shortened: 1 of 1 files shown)


garethfuller (approved):
Nice! Thanks for fixing, it looks like it's fixed the network switch bug too.

PR #78, opened 2024-10-16
feat: swap from pool

agualis:
This can me merged hiding the pool advanced options first.

swapPool

TODO: 
- The swap pool feature is not yet available in the SDK
- Test edge-cases like nested pools


.vscode/settings.json
```diff
@@ -1,7 +0,0 @@
-{
-  "eslint.workingDirectories": [
```
agualis, line 2: @garethfuller deleted this file from git cache as now it is .gitignored
But it would be good to use .editorconfig or similar.

packages/lib/modules/pool/PoolDetail/PoolHeader/PoolHeader.tsx
```diff
@@ -55,14 +56,18 @@ export function PoolHeader() {
...
+              onClick={handleClick}
+              variant="primary"
+              size="lg"
+              isDisabled={isAddLiquidityBlocked}
+              w="full"
+            >
+              Add liquidity
+            </Button>
+            <PoolAdvancedOptions />
+          </HStack>
```
garethfuller, line 70: Just to be clear, you are going to hide this before merging?

packages/lib/modules/swap/useIsPoolSwap.tsx
```diff
@@ -0,0 +1,6 @@
+import { usePathname } from 'next/navigation'
+
+export function useIsPoolSwap() {
+  const pathname = usePathname()
+  return pathname.includes('/pools') && pathname.includes('/swap')
+}
```
garethfuller, line 6: What happens if we allow pool swaps directly in the swap UI eventually?

packages/lib/modules/swap/SwapForm.tsx
```diff
@@ -94,6 +102,22 @@ export function SwapForm() {
...
+      tokens.length === 2 &&
+      tokenSelectKey === 'tokenIn' &&
+      isSameAddress(token.address, tokenOut.address)
+    )
+      return switchTokens()
+    if (
+      tokens.length === 2 &&
+      tokenSelectKey === 'tokenOut' &&
+      isSameAddress(token.address, tokenIn.address)
+    )
```
garethfuller, line 116: Shouldn't this be handled in the same we were handling it in the swap UI? i.e. you just can't select the same token that's currently selected?

garethfuller: Seems like it should be logic that's handled in the modal, not here.

packages/lib/modules/pool/PoolDetail/PoolHeader/PoolHeader.tsx
agualis: Yes

packages/lib/modules/swap/SwapProvider.tsx
```diff
@@ -60,6 +61,8 @@ export type PathParams = {
   amountOut?: string
   // When urlTxHash is present the rest of the params above are not used
   urlTxHash?: Hash
+  // Only used by pool swap
+  poolTokens?: GqlToken[]
```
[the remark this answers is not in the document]
agualis: I refactored the types so it's clearerer that `poolTokens` do not belong to the path

packages/lib/modules/swap/useIsPoolSwap.tsx
agualis: Refactored into `isPoolSwap` + `useIsPoolSwapUrl`

packages/lib/modules/swap/SwapForm.tsx
agualis: I wanted to reuse the same selector that we are using to choose between native asset and wrapped native asset where you can chose the currently selected token.

The logic is handling the case when you chose the other token in a 2-token pool swap. So instead of disabling the other token, it switches them:

https://github.com/user-attachments/assets/259e5c66-1c9a-4abc-a393-975947dc2ea1

packages/lib/modules/swap/useIsPoolSwapUrl.tsx
```diff
@@ -0,0 +1,6 @@
+import { usePathname } from 'next/navigation'
+
+export function useIsPoolSwapUrl() {
+  const pathname = usePathname()
+  return pathname.includes('/pools') && pathname.includes('/swap')
+}
```
garethfuller, line 6: I don't think this is a good approach. I was just trying to point out that the logic using the URL might not be the best way to decide if it's a pool swap. Now, if we decide to support pool swaps in the swap UI you'll have to have two conditionals all over the place.

The alternative approach would be to maintain a `isPoolSwap` state variable in the SwapProvider. The default can be set by a useSwap param (i.e. a param passed into the provider). So the swap layout used in the pool context would take a prop like `isPoolSwap` which is then passed into the provider. Then in the future we can easily switch on this flag with UI changes in the swap form.

garethfuller (changes requested):

packages/lib/modules/swap/useIsPoolSwapUrl.tsx
agualis: We already have the `isPoolSwap` into the provider. 
But we do need the two conditionals. 

`isPoolSwap` is used to display the specific pool swap parts

`useIsPoolSwapUrl` is used to detect the specific case where the swap is happening within a pool actions layout, meaning that receipt page and redirects work in a different way

garethfuller: Ok, maybe let's see.

garethfuller (changes requested):
Hide advanced pool options for production.

PR #79, opened 2024-10-16
feat: Pool list updates for v3

garethfuller:
Updates type column to 'Details' and adds protocol version filter to filters popover. v3 filter option only visible on dev and staging envs.
Screenshot 2024-10-18 at 13 55 18
Screenshot 2024-10-18 at 13 55 25

### DIFF
--- a/packages/lib/modules/pool/PoolList/PoolListFilters.tsx
+++ b/packages/lib/modules/pool/PoolList/PoolListFilters.tsx
@@ -33,7 +33,7 @@ import {
 } from '@chakra-ui/react'
 import { PoolListSearch } from './PoolListSearch'
 import { getProjectConfig } from '@repo/lib/config/getProjectConfig'
-import { usePoolListQueryState } from './usePoolListQueryState'
+import { PROTOCOL_VERSION_TABS, usePoolListQueryState } from './usePoolListQueryState'
 import {
   PoolFilterType,
   poolTypeFilters,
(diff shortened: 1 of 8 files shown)


packages/lib/modules/pool/PoolList/PoolListFilters.tsx
```diff
@@ -380,12 +385,71 @@ const FilterButton = forwardRef<ButtonProps, 'button'>((props, ref) => {
...
+    } else if (poolTypes.includes(GqlPoolType.CowAmm)) {
+      setActiveProtocolVersionTab(tabs[3])
+    } else {
+      setActiveProtocolVersionTab(tabs[0])
+    }
+  }, [])
+
+  useEffect(() => {
+    if (activeProtocolVersionTab.value === 'cow') {
+      togglePoolType(true, GqlPoolType.CowAmm)
```
agualis, line 421: Choosing CowAmm sets pool type but choosing other options should undo this selection.

agualis (changes requested):

packages/lib/modules/pool/PoolList/PoolListFilters.tsx
garethfuller: ahh good catch, I had something in for that initially but then it was causing other issues. Will fix.

agualis (approved):

PR #80, opened 2024-10-16
feat: add lock vebal actions

alter-eggo:
This pr adds lock, unlock, extend lock vebal actions


apps/frontend-v3/app/(app)/pools/[chain]/[variant]/[id]/page.tsx
```diff
@@ -1,10 +1,13 @@
...
 import { TransactionStateProvider } from '@repo/lib/modules/transactions/transaction-steps/TransactionStateProvider'
+import { VebalLockInfoProvider } from '@repo/lib/modules/vebal/lock/VebalLockInfoProvider'
 
 export default function PoolPage() {
   return (
     <TransactionStateProvider>
-      <PoolDetail />
+      <VebalLockInfoProvider>
+        <PoolDetail />
+      </VebalLockInfoProvider>
```
Comment (garethfuller) [inline apps/frontend-v3/app/(app)/pools/[chain]/[variant]/[id]/page.tsx:10]: I'm worried about this being on every pool page. Is it needed on every pool?

packages/lib/modules/pool/actions/PoolActionsLayout.tsx
```diff
@@ -1,49 +1,23 @@
...
-    <Box
-      bg="transparent"
-      left={0}
-      onClick={redirectToPoolPage}
-      pos="absolute"
-      top={0}
-      w="full"
-      zIndex={100}
+    <ModalActionsLayout
+      chain={pool.chain}
```
garethfuller, line 16: 1. `ModalActionsLayout` isn't great naming because it's not really a modal.
   Please rename `FocussedActionLayout` and `FocussedActionNav`.
2. Because you've nested all the logic in `ModalActionsLayout` please get rid of
   `PoolActionsLayout`. This wrapper isn't needed.

packages/lib/modules/tokens/TokensProvider.tsx
```diff
@@ -179,3 +180,15 @@ export function TokensProvider({
...
+  const { getTokensByChain } = useTokens()
+
+  const tokens = getTokensByChain(GqlChain.Mainnet)
+
+  const vebalBptToken = tokens.find(
+    t => t.address === mainnetNetworkConfig.tokens.addresses.veBalBpt
+  )
+
+  return vebalBptToken
+}
```
garethfuller, line 194: Don't create a new hook for this. Just export the vebalBptToken const from the
tokens provider.

packages/lib/shared/services/chakra/themes/base/components.ts
```diff
@@ -798,10 +798,22 @@ export function getComponents(tokens: any, primaryTextColor: string) {
...
+          width: '16px',
+        },
+        track: {
+          _dark: {
+            bg: '#282D34', // fix: what token?
+          },
+          _light: {
+            bg: '#D7D2CB', // inversion of #282D34
+          },
+          height: '5px',
```
garethfuller, line 816: Have you checked how this affects the remove liquidity flow?

packages/lib/shared/hooks/useCurrentDate.ts
garethfuller, line 1: Rename file date.hooks.ts

packages/lib/shared/hooks/useCurrentDate.ts
```diff
@@ -0,0 +1,31 @@
...
+  }, [interval])
+
+  return date
+}
+
+export function useToday(interval = oneMinInMs) {
+  const currentDate = useCurrentDate(interval)
+  const today = startOfDay(currentDate).getTime()
+  return useMemo(() => new Date(today), [today])
+}
```
garethfuller, line 31: Add comment explaining exactly what this hook is returning.

packages/lib/modules/vebal/lock/duration/constants.ts
garethfuller, line 1: Rename lock-duration.constants.ts

packages/lib/modules/vebal/lock/duration/lock-duration.hooks.ts
garethfuller, line 1: Rename useLockDuration.ts

packages/lib/modules/vebal/lock/duration/useLockEndDate.ts
```diff
@@ -0,0 +1,41 @@
...
+  const today = useToday()
+
+  const minLockEndDate = useMemo(() => {
+    return getMinLockEndDate(lockedEndDate ?? today)
+  }, [lockedEndDate, today])
+
+  const maxLockEndDate = useMemo(() => getMaxLockEndDate(today), [today])
+
+  return { minLockEndDate, maxLockEndDate }
+}
```
garethfuller, line 41: Move into lock-duration.hooks.ts

garethfuller: Or just export from useLockDuration.

packages/lib/modules/vebal/lock/duration/useLockEndDate.ts
garethfuller, line 1: Rename lock-duration.utils.ts

packages/lib/modules/vebal/lock/steps/lock.helpers.ts
garethfuller, line 1: Rename lock-step.utils.ts

apps/frontend-v3/app/(app)/pools/[chain]/[variant]/[id]/page.tsx
Comment (garethfuller) [reply apps/frontend-v3/app/(app)/pools/[chain]/[variant]/[id]/page.tsx]: Can you link to the change you made? If you need the boost having a provider or some other solution is not going to change much.

Comment (garethfuller) [reply apps/frontend-v3/app/(app)/pools/[chain]/[variant]/[id]/page.tsx]: Ok, but changing from provider to a normal hook doesn't change anything, so what's the benefit of doing that? It may even introduce problems from using it in multiple places?

What happens with this stale time on the vebal page? What if they just updated their lock information but the data is cached for another minute? Will it result in a UI lag where the user sees their old lock information?

Comment (garethfuller) [reply apps/frontend-v3/app/(app)/pools/[chain]/[variant]/[id]/page.tsx]: Because you use the `useVebalLockData` hook on the pool page, right? So everytime the pool page loads the contents of that hook will be run, basically the same as if it was a provider. The only difference with it being in the provider is that you then can be sure you are only instantiating the hook once, even if you use the hook in multiple places on the pool page.

garethfuller (approved):

PR #85, opened 2024-10-18
chore: fix eslint rules

groninge01:
- fixed eslint rules not found
- configure & run prettier/eslint for both prettier & eslint configs

### DIFF
--- a/packages/prettier-config/index.js
+++ b/packages/prettier-config/index.js
@@ -1,8 +1,3 @@
-module.exports = {
-  singleQuote: true,
-  arrowParens: "avoid",
-  semi: false,
-  proseWrap: "always",
-  printWidth: 100,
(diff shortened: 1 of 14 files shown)


packages/prettier-config/index.js
```diff
@@ -1,8 +1,3 @@
-module.exports = {
-  singleQuote: true,
-  arrowParens: "avoid",
-  semi: false,
-  proseWrap: "always",
-  printWidth: 100,
-  trailingComma: "es5",
-};
+import config from './.prettierrc.json' assert { type: 'json' }
```
groninge01, line 1: this is an experimental option but otherwise we'd need to duplicate the prettier config

garethfuller (approved):

PR #92, opened 2024-10-22
feat: block adding too unbalanced

### LINKED ISSUE #86: Disable single-sided adds
The single asset join does not work when the single asset is > 25% of the pool liquidity.
Detect this case, block single join and show a warning with a clear explanation.

If single asset join AND input > 25% of pool's token balance

Block action
Show warning "You can only add up to 25% of a single asset to the pool in one transaction"

groninge01:
fixes #86 

the issue is not only single sided adds but unbalanced adds in general
the specific error is BAL#304, when this error occurs it is intercepted and a msg is shown

image


packages/lib/modules/pool/actions/add-liquidity/form/AddLiquidityForm.tsx
```diff
@@ -103,6 +107,16 @@ function AddLiquidityMainForm() {
     setPriceImpact(priceImpactQuery.data)
   }, [priceImpactQuery.data])
 
+  useEffect(() => {
+    if (priceImpactQuery.error) {
+      const error = priceImpactQuery.error as PriceImpactErrorType
+      const balError = error.shortMessage.split(/\r?\n/)[1]
```
agualis, line 113: `shortMessage` could be `undefined` for some errors so it's safer to just use `message`
And I would go for `error.message.includes('BAL#304')` which reads better (and performance is not critical here).

agualis (changes requested):
I found some other edge cases testing with big amounts. I will DM you tomorrow.

packages/lib/modules/pool/actions/add-liquidity/AddLiquidityProvider.tsx
```diff
@@ -151,6 +151,7 @@ export function _useAddLiquidity(urlTxHash?: Hash) {
     [!isConnected, LABELS.walletNotConnected],
     [areEmptyAmounts(humanAmountsIn), 'You must specify one or more token amounts'],
     [hasValidationErrors, 'Errors in token inputs'],
+    [needsToUpdateInputs, 'Update inputs first'],
```
garethfuller, line 154: I don't think we should be introducing this as a new state variable. We already have functionality for handling validation type errors. Please try to use the functionality from `TokenInputsValidationProvider` instead. This should then result in the preview being blocked by the `disabledConditions` in the `AddLiquidityProvider`.

garethfuller (changes requested):

packages/lib/modules/pool/actions/add-liquidity/form/AddLiquidityForm.tsx
groninge01: updated!

garethfuller (changes requested):

packages/lib/modules/pool/actions/add-liquidity/form/AddLiquidityForm.tsx
```diff
@@ -103,6 +106,20 @@ function AddLiquidityMainForm() {
     setPriceImpact(priceImpactQuery.data)
   }, [priceImpactQuery.data])
 
+  useEffect(() => {
+    humanAmountsIn.forEach(amount => {
+      if (priceImpactQuery.error?.message.includes('BAL#304') && amount.humanAmount) {
+        setValidationError(amount.tokenAddress, 'Amount is causing an unbalanced join')
```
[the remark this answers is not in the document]
groninge01: Updated!

packages/lib/modules/price-impact/PriceImpactAccordion.tsx
```diff
@@ -88,69 +91,70 @@ export function PriceImpactAccordion({
...
-                  {isUnknownPriceImpact
-                    ? 'Unknown price impact'
-                    : `Price impact is high: Exceeds ${
-                        priceImpactLevel === 'high' ? '1' : '5'
-                      }.00%`}
-                </AlertTitle>
-                <AlertDescription>
-                  <Text color="font.dark" fontSize="sm">
+      {!hasValidationErrors &&
+        (priceImpactLevel === 'high' || priceImpactLevel === 'max' || isUnknownPriceImpact) && (
```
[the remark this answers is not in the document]
groninge01: This change has been reverted.

packages/lib/modules/pool/actions/add-liquidity/form/AddLiquidityForm.tsx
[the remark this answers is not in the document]
garethfuller: I don't think you've pushed the changes?

packages/lib/modules/pool/actions/add-liquidity/AddLiquidityProvider.tsx
groninge01: This has been removed.

packages/lib/modules/pool/actions/add-liquidity/form/AddLiquidityForm.tsx
```diff
@@ -103,6 +105,20 @@ function AddLiquidityMainForm() {
...
+        setValidationError(amount.tokenAddress, 'Amount causes pool imbalance')
+      }
+    })
+  }, [priceImpactQuery.error])
+
+  useEffect(() => {
+    humanAmountsIn.forEach(amount => {
+      setValidationError(amount.tokenAddress, '')
+    })
+  }, [humanAmountsIn])
```
garethfuller, line 120: I'm worried about this overriding a `setValidationError` that is set in the token input. e.g. The user enters more than their balance, what's triggered first? This or the `setValidationError` within the token input? Could this cause a race condition?

packages/lib/shared/utils/error-filters.ts
```diff
@@ -44,3 +44,15 @@ export function isPausedError(error?: Error | null): boolean {
...
+  if (!error) return false
+  return isUnbalancedJoinErrorMessage(error)
+}
+
+export function isUnbalancedJoinErrorMessage(error: Error | null): boolean {
+  const errorStrings = ['BAL#304', 'queryAddLiquidityUnbalanced'] // [v2 error, v3 error]
+  const hasErrors = (errorString: string) => error?.message.includes(errorString)
+
+  return errorStrings.some(hasErrors)
+}
```
garethfuller, line 58: I know it's more verbose, but we're supposed to be using Add or AddLiquidity instead Join terminology.

packages/lib/modules/pool/actions/add-liquidity/form/AddLiquidityForm.tsx
groninge01: Looks like the '>balance' error isn't trigger any more now :(
Maybe remove the token input validation and add back the extra line to `disabledConditions`?
The warning msg is clear about what the user should do.

```
[isUnbalancedAddErrorMessage(priceImpactQuery.error), 'Unbalanced add'],
```

image

image

packages/lib/shared/utils/error-filters.ts
groninge01: Renamed!

packages/lib/modules/pool/actions/add-liquidity/form/AddLiquidityForm.tsx
garethfuller: Yes this is the best solution I think.

garethfuller (approved):

PR #93, opened 2024-10-22
fix: update prettier-config

groninge01:


agualis (approved):
Tested locally that fixed the issue

PR #101, opened 2024-10-24
chore: optimize lint-staged

groninge01:
- add caching for eslint/prettier/typecheck
- set `incremental` to `true` in base tsconfig for typecheck
- -update stylelint and run fix


packages/lib/assets/css/marketing.css
```diff
@@ -103,7 +103,7 @@
   }
 }
 
-@media screen and (min-width >= 768px) {
+@media screen and (width >= 768px) {
```
garethfuller, line 106: Was this intended?

turbo.json
```diff
@@ -41,6 +45,25 @@
       "dependsOn": ["^graphql:gen"],
       "inputs": [".env*"],
       "outputs": ["shared/services/api/generated/**"]
+    },
+    "transit": {
+      "dependsOn": ["^transit"]
+    },
+    "//#format": {
+      "outputs": ["node_modules/.cache/.prettiercache"]
+    },
```
garethfuller, line 54: Where did this solution come from? Can you point me to the docs?

package.json
```diff
@@ -8,10 +8,12 @@
...
     "dev:beets": "turbo dev --filter beets-frontend-v3",
     "graphql:gen": "turbo graphql:gen",
+    "lint:all": "turbo run lint format --continue",
+    "lint:all:fix": "turbo run lint:fix format:fix --continue --force",
     "lint": "turbo lint",
     "lint:fix": "turbo lint:fix",
-    "prettier": "prettier --check '**/*.*(md|json|yaml|ts|js|tsx)' --log-level warn",
-    "prettier:fix": "pnpm run prettier --write",
+    "format": "prettier '*.*(md|json|yaml|ts|js|tsx)' --check --cache --cache-location='node_modules/.cache/.prettiercache'",
+    "format:fix": "prettier '*.*(md|json|yaml|ts|js|tsx)' --write --cache --cache-location='node_modules/.cache/.prettiercache' --log-level=warn",
```
garethfuller, line 16: Should this not be?

```suggestion
    "format": "turbo format",
    "format:fix": "turbo format:fix",
```

packages/lib/assets/css/marketing.css
groninge01: stylelint threw an error on it

```
> @balancer/frontend-monorepo@ stylelint /home/groninge/projects/frontend-monorepo
> stylelint '**/*.css'


packages/lib/assets/css/marketing.css
  106:18  ✖  Unexpected invalid media query " (min-width >= 768px)", "min-" and "max-" prefixes are not needed when using range queries  media-query-no-invalid
```

turbo.json
groninge01: https://prettier.io/blog/2023/11/30/cli-deep-dive#caching

package.json
groninge01: this is needed to run prettier in the root (and only in the root)

packages/lib/assets/css/marketing.css
garethfuller: @uiuxxx is this linting change valid? or will it cause problems.

turbo.json
garethfuller: But what is this `transit` and `//#format` command? I've never seen that in Turbo.

groninge01: The task with double slashes is [a root task](https://turbo.build/repo/docs/crafting-your-repository/configuring-tasks#registering-root-tasks). 
[The `transit` task](https://turbo.build/repo/docs/crafting-your-repository/configuring-tasks#dependent-tasks-that-can-be-ran-in-parallel) is turbo magic to make tasks run in parallel (but I only saw that happen once).

garethfuller (approved):
I'm a bit worried this is overcomplicating things. Would be great to look into biome.

PR #112, opened 2024-10-29
fix: add pool chain to swap state when isPoolSwap

agualis:
Also blocks pool swaps when there's no token metadata.


packages/lib/modules/swap/SwapProvider.tsx
```diff
@@ -158,6 +158,9 @@ export function _useSwap({ poolActionableTokens, pool, pathParams }: SwapProvide
 
   if ((isTokenInSet && !tokenInInfo) || (isTokenOutSet && !tokenOutInfo)) {
     try {
+      if (isPoolSwap) {
```
agualis, line 161: Maybe we need a special UI alert for this case but I will add it after deploying the live fix.

groninge01 (approved):

PR #126, opened 2024-10-31
feat: Hooks on pool page

garethfuller:
- [x] Hook icon in header
- [x] Hook contract address in contracts section

Screenshot 2024-10-31 at 16 18 26
Screenshot 2024-10-31 at 16 18 30

### DIFF
--- a/packages/lib/modules/hooks/useHook.ts
+++ b/packages/lib/modules/hooks/useHook.ts
@@ -1,0 +1,9 @@
+import { Pool } from '../pool/PoolProvider'
+
+export function useHook(pool: Pool) {
+  const hasHook = pool.hook?.address
+
+  return {
+    hasHook,
+  }
+}
--- a/packages/lib/modules/hooks/useHook.ts
+++ b/packages/lib/modules/hooks/useHook.ts
@@ -1,9 +1,13 @@
+import { Address } from 'viem'
 import { Pool } from '../pool/PoolProvider'
 
 export function useHook(pool: Pool) {
-  const hasHook = pool.hook?.address
+  const hasHook = !!pool.hook?.address
+
+  const hookAddress = (pool.hook?.address || '0x') as Address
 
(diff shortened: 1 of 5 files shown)


packages/lib/modules/pool/PoolDetail/PoolHeader/PoolMetaBadges.tsx
```diff
@@ -46,6 +50,13 @@ export default function PoolMetaBadges() {
       <BalBadge color="font.secondary" fontSize="xs" textTransform="none">
         {getPoolTypeLabel(pool.type)}
       </BalBadge>
+      {hasHook && (
+        <BalBadge color="font.primary" fontSize="xs" h={8} w={8}>
```
agualis, line 54: I would add a popover like we have in the swap fee icon: 

hookTooltip

If it's easy we could show a link to the name of the hook or a link to the hook contract in the popover. (can be done in the future)

packages/lib/modules/hooks/useHook.ts
```diff
@@ -0,0 +1,13 @@
+import { Address } from 'viem'
+import { Pool } from '../pool/PoolProvider'
+
+export function useHook(pool: Pool) {
+  const hasHook = !!pool.hook?.address
```
agualis, line 5: Out of curiosity: cannot we have more than one hook in a pool?

packages/lib/modules/pool/PoolDetail/PoolHeader/PoolMetaBadges.tsx
garethfuller: Yeah metadata is coming in another PR.

packages/lib/modules/hooks/useHook.ts
garethfuller: Nope.

PR #138, opened 2024-11-02
chore: add sign icon or gas icon to step buttons

agualis:
Examples: 

permit2

remove


packages/lib/shared/components/btns/button-group/LabelWithIcon.tsx
```diff
@@ -0,0 +1,21 @@
+import { HStack, Text } from '@chakra-ui/react'
+import { PropsWithChildren } from 'react'
+import { SignIcon } from '../../icons/SignIcon'
+import { GasIcon } from '../../icons/GasIcon'
+
+type Props = {
```
agualis, line 6: We could pass the icon as a ReactNode to make it more flexible but keeping it simple for now.

uiuxxx (approved):
Nice!

PR #141, opened 2024-11-04
feat: nested pool display

### LINKED ISSUE #99: Nested pool display
- [ ] Pool detail
- [ ] Pool list

groninge01:
fixes #99 

PoolDetail:

- indent tokens of nested pool(s)
- show tokens of nested pool(s) in weight chart (instead of nested bpt(s) )
- show legend below weight chart
- add nested pools risk link (@uiuxxx this needs copy on `.../risks#nested-pools`)
- update pool token pills (also for PoolList)

PoolList:

- ~~no design so no work needed?~~


packages/lib/modules/tokens/TokensProvider.tsx
```diff
@@ -110,6 +110,14 @@ export function _useTokens(
     return price.price
   }
 
+  // this also fetches the price for a bpt
+  function priceForAddress(address: string, chain: GqlChain): number {
+    const price = getPricesForChain(chain).find(price => isSameAddress(price.address, address))
+    if (!price) return 0
+
+    return price.price
+  }
```
garethfuller, line 119: What's the point in this function instead of just using `priceFor`? https://github.com/balancer/frontend-monorepo/blob/ca1f61bd9c948d8c9d5f637ebed644b42099ae95/packages/lib/modules/tokens/TokensProvider.tsx#L119

Is it just that getToken doesn't work for BPT? If so, why not adapt `priceFor` to fallback to this logic if `token` isn't found?

garethfuller (commented):
Nested token amounts and weights should have a lighter font. See design, it's quite subtle.
Screenshot 2024-11-07 at 15 50 11

Screenshot 2024-11-07 at 15 52 48

garethfuller (commented):
Instead of the BPT token we should be showing the underlying token icons here I think. Same in the pool list.
Screenshot 2024-11-07 at 15 50 39

Screenshot 2024-11-07 at 15 54 06

packages/lib/modules/tokens/TokensProvider.tsx
groninge01: Tokens does not contain BPTs. I have updated `priceFor` with the fallback.

groninge01:
> Nested token amounts and weights should have a lighter font. See design, it's quite subtle. 

updated with values from design

image

groninge01:
> Instead of the BPT token we should be showing the underlying token icons here I think. Same in the pool list. 

added support for boosted to token pills 

note: api still needs to expose underlying tokens in `displayTokens`

groninge01:
converted to draft again because of api updates not yet available in prod

groninge01:
can be reviewed, api needs to deploy before merge

groninge01:
> All checks are failing here, can you fix?

api is synced and merge conflicts are solved, all green now

groninge01:
> empty hook string https://mono-test-v3-git-feat-nested-pool-display-balancer.vercel.app/pools/sepolia/v3/0x711fd80b36723bce3b42ad6622903e1e39d911dd
> 
> Screenshot 2024-11-25 at 16 36 09

fixed!

groninge01:
> hm, I know it's not about this pr, but seems hook icon lost from pool page header? may be some problems when we cherry picked commits from previous pr? also icon seems cut
> 
> btw I checked and hook is null [here](https://mono-test-v3-git-feat-nested-pool-display-balancer.vercel.app/pools/sepolia/v3/0x8eb3c3e484d5163881d7499cbe68aad531993479) Screenshot 2024-11-25 at 16 47 32

yeah i saw the icon was cut off and looked briefly into it but couldn't find the issue yet
as it is a hook in a nested pool (which probs won't be launched in prod yet) I left it for now
same goes for the hook badge in the pool header, I didn't add nested pool hook support there yet (but left a TODO comment for when we do need it)

packages/lib/modules/pool/PoolDetail/PoolInfo/PoolContracts.tsx
garethfuller, line 1: Fix unused import

garethfuller (commented):
Deployments are failing?

garethfuller (approved):

PR #144, opened 2024-11-06
fix: lock file

alter-eggo:
This pr fixed pnpm lock file


garethfuller:
I think this is fixed in #145, the lockfile change is less extreme there.

PR #145, opened 2024-11-06
fix: Lockfile / checks

garethfuller:
Also includes some minor updates to links that pointed to the old repo.


garethfuller:
Had to bypass husky to create this branch. Typecheck doesn't work for me locally.

PR #148, opened 2024-11-07
feat: voting list

alter-eggo:
This pr adds vebal voting list
Screenshot 2024-11-07 at 13 34 50


packages/lib/modules/vebal/vote/hidden-hand/hidden-hand.type.ts
garethfuller, line 1: hidden-hand.types.ts

packages/lib/modules/vebal/vote/gauge/useGaugeVotes.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const { userAddress, isConnected } = useUserAccount()
+
+  const thisWeekTimestamp = useMemo(() => {
+    // this makes sure we don't compute votes from before Mar31 in the "This period" entry,
+    // since the system is not fully active between Mar31 and Apr6
+    // (ie the first period starts on Apr7)
+    const thisWeekTimestamp = toUnixTimestamp(Math.floor(Date.now() / oneWeekInMs) * oneWeekInMs)
+    if (thisWeekTimestamp == FIRST_WEEK_TIMESTAMP) {
+      return thisWeekTimestamp - oneWeekInSecs
+    }
```
garethfuller, line 62: Is this not legacy code from launching vebal? Why would we need it anymore?

packages/lib/modules/vebal/vote/VoteRateTooltip.tsx
```diff
@@ -0,0 +1,162 @@
...
+    <VoteDownIcon />
+  )
+
+  /* fix: where is data? */
+  const voteState = useMemo(() => {
+    return {
+      currentPeriodVebal: Math.random() * 20000,
+      nextPeriodVebal: Math.random() * 20000,
+    }
+  }, [])
```
garethfuller, line 105: You need to figure this out, but we can try and get the PR merged without it for now. However, please don't fill they UI with data that looks like it could be real, otherwise we might eventually forget and mistake it for real data. Please put a dash or something that makes it obvious in the UI that this still needs to be done.

garethfuller: Also, why does something like this need to be a useMemo? It's just a static const.

packages/lib/modules/vebal/vote/VoteList/VoteListProvider.tsx
```diff
@@ -0,0 +1,89 @@
...
+      ? previousData.veBalGetVotingList
+      : data?.veBalGetVotingList || []
+
+  const pagination = queryState.pagination
+
+  const {
+    data: votingIncentives,
+    isLoading: votingIncentivesLoading,
+    error: votingIncentivesError,
+  } = useHiddenHandVotingIncentives()
```
garethfuller, line 36: Any fetching that can be done server-side, e.g to external APIs or our own can be done in the RSC/layout component and passed into providers as inputs. That way the data is server side rendered and it speeds everything up.

apps/frontend-v3/app/(app)/vebal/vote/layout.tsx
garethfuller, line 1: You should be fetching data in this component and passing it into providers.

garethfuller (changes requested):

agualis (changes requested):
Minor style issue:
Screenshot 2024-11-21 at 13 50 06

the botton of the two gauge "g"s is cut by the table.

packages/lib/modules/vebal/vote/gauge/useGaugeVotes.tsx
```diff
@@ -0,0 +1,134 @@
...
+        id: `${gaugeAddress}.${v.path}`,
+        abi: AbiMap['balancer.gaugeControllerAbi'] as any,
+        address: mainnetNetworkConfig.contracts.gaugeController as Hex,
+        functionName: v.fn,
+        args: v.args,
+      }))
+    })
+  }, [gaugeAddresses, thisWeekTimestamp, nextWeekTimestamp, isConnected, userAddress])
+
+  const { results, refetchAll, isLoading } = useMulticall(gaugeRequests, {
```
agualis, line 97: In general, I think that we should try to avoid `useMulticall` as much as possible and replace it with `useReadContracts`, which offers better TS inference experience and it will be easier to refactor in the future if wagmi changes. 

`useMulticall` was a workaround at the beginning of the project, but it lacks of good TS experience cause it uses:
`.reduce((o, { id, ...rest }) => set(o, id, rest), {})` which is not TS safe.

There are already some examples of complex queries using `useReadContracts` in the code base.

agualis (changes requested):

packages/lib/modules/vebal/vote/gauge/useGaugeVotes.tsx
garethfuller, line 1: This doesn't need to be in it's own folder.

garethfuller (changes requested):
The votes don't seem to be working anymore.

Screenshot 2025-01-08 at 16 08 25

packages/lib/modules/vebal/vote/VoteList/VoteList.tsx
```diff
@@ -0,0 +1,48 @@
...
+        .catch(err => [undefined, err] as const),
+    ])
+
+  return (
+    <VoteListProvider
+      data={voteListQueryData}
+      // Only plain objects are allowed as props within RSC
+      error={errorToJson(error)}
+      votingIncentives={votingIncentives}
+      votingIncentivesError={errorToJson(votingIncentivesError)}
```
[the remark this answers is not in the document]
garethfuller: I noticed in the next level down that you do use error boundaries, but for some reason this still seems bad. Can you check NextJS docs to see if they have any recommendations for this kind of thing?

apps/frontend-v3/app/(app)/vebal/manage/layout.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { TransactionStateProvider } from '@repo/lib/modules/transactions/transaction-steps/TransactionStateProvider'
+
+export default function VeBALManageLayout({ children }: PropsWithChildren) {
+  const { vebalBptToken } = useTokens()
+
+  if (!vebalBptToken) throw new Error('vebalBptToken not found')
+
+  return (
+    <TokenBalancesProvider initTokens={[vebalBptToken]}>
+      <CrossChainSyncProvider>
```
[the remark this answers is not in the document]
alter-eggo: yeah, I've put provider  here to avoid unnecessary cross chain sync related queries in the app
later when feature is ready I'll put it higher

packages/lib/modules/vebal/vote/VoteList/VoteList.tsx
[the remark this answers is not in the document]
alter-eggo: removed error
removed votingIncentivesError

we cannot use error boundary for hh request since we should render table even if this request fails

so now I just pass votingIncentivesErrorMessage, I will need to use in table to show the reason why data is not there

garethfuller (changes requested):
Votes still don't seem to be working in the table...

garethfuller (approved):

PR #152, opened 2024-11-07
feat: proportional v3 boosted adds

agualis:
Note that there is an unresolved issue: https://github.com/balancer/frontend-monorepo/pull/153

Adds v3 proportional boosted: 

https://github.com/user-attachments/assets/9a718cb2-c4ce-4104-9abb-166e80fb1812


packages/lib/modules/pool/actions/LiquidityActionHelpers.ts
```diff
@@ -103,7 +106,39 @@ export class LiquidityActionHelpers {
...
+  public get boostedPoolStateWithBalances(): PoolStateWithBalances {
+    const underlyingTokensWithBalance: PoolTokenWithBalance[] = this.pool.poolTokens.map(
+      (token, index) =>
+        token.underlyingToken
+          ? {
+              address: token.underlyingToken?.address as Address,
+              decimals: token.underlyingToken?.decimals as number,
+              index,
+              // TODO: balance: token.underlyingToken?.balance * rate as HumanAmount,
+              balance: token.balance as HumanAmount,
```
agualis, line 124: @garethfuller Note that this balance should be the underlying balance (wrappingToken.balance * priceRate) instead of just wrappingToken balance but I'm investigating why that's failing.
Related PR on top of this one: 
https://github.com/balancer/frontend-monorepo/pull/153

PR #153, opened 2024-11-07
chore: use priceRate * balance

agualis:


garethfuller (commented):
And this works now? I thought it was failing when you tried this before?

groninge01 (approved):

PR #154, opened 2024-11-08
feat: v3 nested boosted adds

agualis:
Test with nested reference pool:
https://mono-test-v3-git-feat-nestedboostedaddsremoves-balancer.vercel.app/pools/sepolia/v3/0x0270daf4ee12ccb1abc8aa365054eecb1b7f4f6b

https://github.com/user-attachments/assets/06c9ebd8-7184-43f0-8ef8-60c4b098c991


garethfuller:
2% price impact for an add like that seems extreme. What's the pool link so I can check the balances? Also, looks like the price impact was actually positive for that add? Maybe we are showing positive price impacts as negative instead of zero?

agualis:
> 2% price impact for an add like that seems extreme. What's the pool link so I can check the balances? Also, looks like the price impact was actually positive for that add? Maybe we are showing positive price impacts as negative instead of zero?

See PR description:
https://mono-test-v3-git-feat-nestedboostedaddsremoves-balancer.vercel.app/pools/sepolia/v3/0x0270daf4ee12ccb1abc8aa365054eecb1b7f4f6b

package.json
```diff
@@ -17,6 +17,8 @@
     "stylelint": "stylelint '**/*.css'",
     "stylelint:fix": "pnpm run stylelint -- --fix",
     "test:integration": "turbo test:integration --log-order=stream --",
+    "test:unit:lib": "pnpm --filter lib run test:unit",
```
agualis, line 20: Exposes lib test scripts without turbo to improve DX

garethfuller: This goes against the Turbo patterns. If it's the terminal UI that is the problem please look into using the UI flag instead: https://turbo.build/repo/docs/reference/run#--ui

packages/lib/modules/pool/actions/LiquidityActionHelpers.ts
garethfuller, line 1: Fix linting issue.

packages/lib/modules/tokens/approvals/permit2/signPermit2Add.tsx
```diff
@@ -62,17 +62,26 @@ async function sign({
...
     chain: pool.chain,
     amountsIn: sdkQueryOutput.amountsIn,
   })
 
-  const signFn = isBoosted(pool)
-    ? Permit2Helper.signAddLiquidityBoostedApproval
-    : Permit2Helper.signAddLiquidityApproval
+  function getSignFn() {
+    if (isV3WithNestedActionsPool(pool)) {
+      console.log('isV3WithNestedActionsPool edge case')
```
garethfuller, line 73: Remove?

package.json
agualis: Updated.

package.json
```diff
@@ -17,7 +17,9 @@
     "stylelint": "stylelint '**/*.css'",
     "stylelint:fix": "pnpm run stylelint -- --fix",
     "test:integration": "turbo test:integration --log-order=stream --",
+    "test:integration:stream": "turbo test:integration --ui=stream --",
     "test:unit": "turbo test:unit --log-order=stream --",
+    "test:unit:stream": "turbo test:unit --ui=stream --",
```
garethfuller, line 22: If you find it more useful to use stream here, just update the existing commands. Likely everyone would rather what you prefer.

agualis: I prefer to keep both for a period of time until I choose one or the other.

garethfuller: nah because we'll never get rid of it in the future. I'd rather we keep it clean. Just change the original commands. If someone complains then we'll know if we need more commands.

agualis: Ok, merged options.

garethfuller (approved):

PR #166, opened 2024-11-18
fix: undefined error on swap page

alter-eggo:
Fix unhandled error on swap page, [discord link](https://discord.com/channels/638460494168064021/638529669330894868/1306973621687812179)

I realise this is not the best fix approach but there is general data handling problem in the app, when we pass possible undefined values in the hooks which leads to such errors and necessity to check for undefined in each nested hook. 
I believe "right" fix would lead to huge refactoring

### DIFF
--- a/packages/lib/modules/swap/usePermit2SwapStep.tsx
+++ b/packages/lib/modules/swap/usePermit2SwapStep.tsx
@@ -30,7 +30,7 @@ export function useSignPermit2SwapStep({
   const { userAddress } = useUserAccount()
   const { slippage } = useUserSettings()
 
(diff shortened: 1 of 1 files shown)


groninge01:
maybe just disable the 'switch' button when one of the inputs is empty?

```
isDisabled={tokenIn.address === emptyAddress || tokenOut.address === emptyAddress}
```

groninge01:
> @groninge01 good idea, but we save swap state, so once user tried to switch, he won't be able to reach swap page...

yeah those users need to delete swapState (once) from local storage

groninge01 (approved):

PR #171, opened 2024-11-20
 feat: add erc4626 review data

### LINKED ISSUE #161: erc4626 review support
If a pool has erc4626 (boosted) tokens we need to handle the provided reviews like hooks and rate providers.

groninge01:
fixes #161


apps/beets-frontend-v3/package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "build": "pnpm --filter @repo/lib graphql:gen && next build",
     "dev": "concurrently \"pnpm --filter @repo/lib graphql:gen --watch\" \"next dev -p 3001\"",
+    "dev:mock": "next dev -p 3001",
```
agualis, line 9: I guess that you use this locally with your mock setup but the name of the script could be confusing for someone external. Maybe better calling it something like `dev:without-graphql-gen` or `dev:no-gen`?

groninge01: Renamed them

agualis (approved):

PR #175, opened 2024-11-21
feat: update pool detail for beets

groninge01:


groninge01:
reverted moving the veballockprovider as that was causing errors on the detail (and portfolio) page

agualis (approved):

packages/lib/modules/pool/actions/add-liquidity/modal/StakingOptions.tsx
```diff
@@ -54,32 +62,34 @@ export function StakingOptions() {
...
-                  isOpen={auraDisclosure.isOpen}
-                  onClose={auraDisclosure.onClose}
-                  partner={RedirectPartner.Aura}
-                  redirectUrl={getAuraPoolLink(getChainId(chain), pool.staking.aura.auraPoolId)}
-                />
-              </>
-            )}
-          </VStack>
-        </Card>
+        {isBalancerProject() && pool.staking?.aura && (
```
garethfuller, line 65: I'm a bit worried about conditional rendering like this within the lib folder. I guess somehow we need to split components like this out into sub-project-specific packages. But I realise you're in a rush so we can do this for now.

packages/lib/config/getProjectConfig.ts
```diff
@@ -14,3 +14,11 @@ const PROJECT_CONFIG = process.env.NEXT_PUBLIC_PROJECT_ID
...
   return PROJECT_CONFIG
 }
+
+export function isBeetsProject() {
+  return getProjectConfig().projectId === 'beets'
+}
+
+export function isBalancerProject() {
+  return getProjectConfig().projectId === 'balancer'
+}
```
garethfuller, line 24: Can these not just be consts? 
```suggestion
export const isBeetsProject = getProjectConfig().projectId === 'beets'
export const isBalancerProject = getProjectConfig().projectId === 'balancer'
```

packages/lib/modules/pool/actions/add-liquidity/modal/StakingOptions.tsx
```diff
@@ -24,13 +25,15 @@ export function StakingOptions() {
 
   const auraDisclosure = useDisclosure()
 
+  const projectConfig = getProjectConfig()
```
garethfuller, line 28: Same with this, can this not just be a global const?

garethfuller (approved):
Approved to unblock, but please address comments where relevant.

packages/lib/config/getProjectConfig.ts
groninge01: yeah when you also implement the comment below -> updated

packages/lib/modules/pool/actions/add-liquidity/modal/StakingOptions.tsx
groninge01: added but i haven't changed it everywhere yet -> #325

PR #184, opened 2024-11-26
feat: remove pool swap handler

alter-eggo:
This pr removes pool swap handler, since we can get swap paths from sor

### DIFF
--- a/packages/lib/modules/swap/handlers/DefaultSwap.handler.integration.spec.tsx
+++ b/packages/lib/modules/swap/handlers/DefaultSwap.handler.integration.spec.tsx
@@ -1,0 +1,91 @@
+/* eslint-disable max-len */
+import { ExactInQueryOutput, ExactOutQueryOutput } from '@balancer/sdk'
+import mainnetNetworkConfig from '@repo/lib/config/networks/mainnet'
+import { daiAddress, wETHAddress } from '@repo/lib/debug-helpers'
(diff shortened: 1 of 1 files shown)


packages/lib/modules/swap/SwapProvider.tsx
```diff
@@ -199,6 +205,7 @@ export function _useSwap({ poolActionableTokens, pool, pathParams }: SwapProvide
       tokenOut: swapState.tokenOut.address,
       swapType: swapState.swapType,
       swapAmount: getSwapAmount(),
+      poolIds: getPoolIds(),
```
agualis, line 208: Please add a comment to make the filter usage explicit. For example: 
"We only use this field to filter by the specific pool swap path when swapping from a concrete pool"

agualis (changes requested):

packages/lib/modules/swap/SwapProvider.tsx
```diff
@@ -87,9 +85,9 @@ function selectSwapHandler(
     return new AuraBalSwapHandler(tokens)
   }
 
-  if (pool && poolActionableTokens) {
-    return new PoolSwapHandler(pool, poolActionableTokens)
-  }
+  // if (pool && poolActionableTokens) {
```
agualis, line 88: Once you delete this, it is possible that you can remove pool and poolActionableTokens params from `selectSwapHandler` function.

agualis (approved):

PR #209, opened 2024-11-28
fix: return zero price impact when user wantsProportional

agualis:
Now adding in proportional mode will return hadcoded zero price impact to avoid problems in PI calculations. 

https://mono-test-v3-git-fix-proportionalzeropriceimpact-balancer.vercel.app/pools/sepolia/v3/0x2ff3b96e0057a1f25f1d62ab800554ccdb268ab8/add-liquidity

Screenshot 2024-11-28 at 16 44 54


agualis:
@groninge01 if the user uses the same input amounts but its not in "wantsProportional mode" they will see this Unknown price impact error, which I consider bad UX. Should we enforce "wantsProportional mode" when the liquidity in the pool is low?

Screenshot 2024-11-28 at 16 52 11

PR #211, opened 2024-11-28
feat: add hook tag on pools list

alter-eggo:
This pr adds hook tag on pools list
Screenshot 2024-11-28 at 21 31 44

### DIFF
--- a/packages/lib/modules/hooks/useHook.ts
+++ b/packages/lib/modules/hooks/useHook.ts
@@ -1,30 +1,19 @@
-import {
-  GqlChain,
-  GqlHook,
-  GqlPoolTokenDetail,
-} from '@repo/lib/shared/services/api/generated/graphql'
 import { useHooks } from './HooksProvider'
 import { getChainId } from '@repo/lib/config/app.config'
+import { Pool } from '../pool/PoolProvider'
+import { PoolListItem } from '../pool/pool.types'
 
-export function useHook({
-  hook,
-  poolTokens,
-  chain,
-}: {
-  hook: GqlHook | null | undefined
-  poolTokens: GqlPoolTokenDetail[]
-  chain: GqlChain
-}) {
+export function useHook(pool: Pool | PoolListItem) {
   const { metadata } = useHooks()
-  const hasHook = !!hook?.address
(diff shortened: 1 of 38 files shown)


packages/lib/modules/hooks/useHook.ts
```diff
@@ -1,18 +1,30 @@
...
+} from '@repo/lib/shared/services/api/generated/graphql'
 import { useHooks } from './HooksProvider'
 import { getChainId } from '@repo/lib/config/app.config'
 
-export function useHook(pool: Pool) {
+export function useHook({
+  hook,
+  poolTokens,
+  chain,
+}: {
```
garethfuller, line 13: Please revert to just pass in the pool. It's more verbose in all the components you need to use this hook or the hook tag component otherwise.

garethfuller (dismissed):

packages/lib/modules/hooks/useHook.ts
alter-eggo: There are different pool types in pools vs pool query, that's why decided to pass just properties

garethfuller: Yeah, like I said, can you not just use the type `Pool | PoolListItem`?

alter-eggo: fixed

packages/lib/modules/pool/PoolList/PoolListTable/PoolListTableRow.tsx
```diff
@@ -15,6 +15,8 @@ import { isBoosted } from '../../pool.helpers'
 import { TokenIcon } from '@repo/lib/modules/tokens/TokenIcon'
 import { useErc4626Metadata } from '../../../erc4626/Erc4626MetadataProvider'
 import Image from 'next/image'
+import { PoolHookTag } from '../../PoolDetail/PoolHookTag'
+import { GqlPoolTokenDetail } from '@repo/lib/shared/services/api/generated/graphql'
```
agualis, line 19: Delete unused import

agualis (changes requested):

agualis (approved):

PR #212, opened 2024-11-29
feat: handle v3 invariant add errors

agualis:
Note that we will display the alert and "advice" the user to use proportional inputs in this cases: 
- v3 vault invariant errors
- v2 unbalanced error 

https://github.com/user-attachments/assets/55f844ff-f439-43ee-b8ed-256cb04d7f65

This PR also fixes a GraphQL field  deprecation (`queryBatchSwap` field)

### DIFF
--- a/packages/lib/shared/utils/error-filters.ts
+++ b/packages/lib/shared/utils/error-filters.ts
@@ -65,7 +65,10 @@ export function isUnbalancedAddErrorMessage(error: Error | null): boolean {
 }
 
 export function isInvariantRatioSimulationErrorMessage(errorMessage?: string): boolean {
-  return errorMessage?.includes('InvariantRatioAboveMax') || errorMessage?.includes('InvariantRatioBelowMin') 
+  return (
+    !!errorMessage?.includes('InvariantRatioAboveMax') ||
+    !!errorMessage?.includes('InvariantRatioBelowMin')
+  )
 }
 
 export function isInvariantRatioPIErrorMessage(errorMessage?: string): boolean {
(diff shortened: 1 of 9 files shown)


packages/lib/shared/utils/error-filters.ts
```diff
@@ -56,3 +63,20 @@ export function isUnbalancedAddErrorMessage(error: Error | null): boolean {
 
   return errorStrings.some(hasErrors)
 }
+
+export function isInvariantRatioSimulationErrorMessage(errorMessage?: string): boolean {
+  if (errorMessage?.includes('InvariantRatioAboveMax')) return true
+  if (errorMessage?.includes('InvariantRatioBelowMin')) return true
```
groninge01, line 69: use or?

packages/lib/shared/utils/query-errors.ts
```diff
@@ -355,6 +356,17 @@ export function shouldIgnore(message: string, stackTrace = ''): boolean {
 
   if (isPausedErrorMessage(message)) return true
 
+  /*
+    When hitting Invariant Ratio Above max error (only in v3 pools) we enforce proportional UX so we don't need sentry logs
+    Context: https://github.com/balancer/balancer-maths/blob/8aaf871acd9e138ba855f03be723cdfd630f4246/typescript/src/weighted/weightedMath.ts#L10
+    */
+  if (
+    isInvariantRatioSimulationErrorMessage(message) ||
+    isInvariantRatioSimulationErrorMessage(message)
```
groninge01, line 365: double check?

packages/lib/shared/utils/error-filters.ts
agualis: I don't have an strong opinion unless the options would grow.
Would you prefer `or` or this shorter version: 
`['InvariantRatioAboveMax', 'InvariantRatioBelowMin'].includes(errorMessage)` ?

groninge01 (approved):

PR #214, opened 2024-11-29
feat: add tabs for add liquidity

### LINKED ISSUE #149: All v3 pools threshold force proportional
At certain threshold of added liquidity to pool balance, unbalanced adds will fail and the user will need to switch over to a forced proportional add type.

In order to make the UX deterministic we are going to refactor proportional UX implementation for all pools:

We will have 2 tabs in the add liquidity form:
- 1 tab for unbalanced adds  
- 1 tab for proportional adds
unbalanced is the default one but there are flows when it will be disabled, for example, when `requiresProportionalInput` (CoW, Gyro,…).

If the user hits the MaxRatio threshold in v3 pools, we will show an alert with a proper context and a call to action to move to proportional adds tab.

We also want a tooltip info icon to give context on hover when the unbalanced tab is disabled
For example, instead of the current alert for `requiresProportionalInput` cases:
Image

We will show the info tooltip close to the tabs with this tooltip:  “this pool requires liquidity to be added proportionally”

There will be an edge cases when we will enforce proportional mode (disabling unbalanced tab): 
- when the pool has TVL < USD50000
in that case the tooltip will say “liquidity must be added proportionally until the pool TVL is greater than $50000”
(that corner case avoids unknown price impact errors when the pool is in its initial stage with low liquidity)

Note that this $50000 threshold is just an initial attempt and we could make it dynamic if we discover that its useful.

The implementation of the "proportional tab mode" will also change in all pools: 
We used to have UnblancedAdd kind + calculateProportional inputs. 
Now we will always use explicit ProportionalAdd kind with exact out input. This will simplify code and make the UX more deterministic.

groninge01:
fixes #149 

- Missing integration tests for the new Proportional Handler.


packages/lib/modules/pool/actions/add-liquidity/AddLiquidityProvider.tsx
```diff
@@ -91,6 +93,19 @@ export function _useAddLiquidity(urlTxHash?: Hash) {
     ])
   }
 
+  function clearAmountsIn(changedAmount?: HumanTokenAmountWithAddress) {
```
agualis, line 96: i moved this function from useProportionalInputs to here

agualis (approved):

PR #215, opened 2024-11-29
refactor: real proportional adds

agualis:
Refactor to use AddLiquidityKind.Proportional (instead of AddLiquidityKind.Unbalanced + calculated proportional amounts) whenever is possible (some v2 stable pools do not support it).

### DIFF
--- a/packages/lib/modules/pool/actions/LiquidityActionHelpers.ts
+++ b/packages/lib/modules/pool/actions/LiquidityActionHelpers.ts
@@ -40,6 +40,7 @@ import {
   isComposableStableV1,
   isCowAmmPool,
   isGyro,
+  isStable,
   isUnbalancedLiquidityDisabled,
   isV3Pool,
   isV3WithNestedActionsPool,
@@ -275,6 +276,11 @@ export function shouldUseRecoveryRemoveLiquidity(pool: Pool): boolean {
 export function requiresProportionalInput(pool: Pool): boolean {
   if (isV3Pool(pool) && isUnbalancedLiquidityDisabled(pool)) return true
   return isGyro(pool.type) || isCowAmmPool(pool.type)
+}
+
+// Some pool types do not support AddLiquidityKind.Proportional in the SDK
+export function supportsProportionalAddLiquidityKind(pool: Pool): boolean {
+  return !isStable(pool.type)
 }
 
 type ProtocolVersion = PoolState['protocolVersion']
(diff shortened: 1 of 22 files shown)


packages/lib/modules/pool/actions/LiquidityActionHelpers.ts
```diff
@@ -277,6 +278,17 @@ export function requiresProportionalInput(pool: Pool): boolean {
   return isGyro(pool.type) || isCowAmmPool(pool.type)
 }
 
+// Some pool types do not support AddLiquidityKind.Proportional in the SDK
+export function supportsProportionalAddLiquidityKind(pool: Pool): boolean {
+  if (
+    isV2Pool(pool) &&
```
agualis, line 284: I prefer to add this explicit clause guard instead of a single return because we will have more edge cases (maybe YAGNI but it's ok)

garethfuller (approved):

packages/lib/modules/pool/actions/add-liquidity/AddLiquidityProvider.tsx
```diff
@@ -41,6 +41,8 @@ export function _useAddLiquidity(urlTxHash?: Hash) {
   const [acceptPoolRisks, setAcceptPoolRisks] = useState(false)
   const [wethIsEth, setWethIsEth] = useState(false)
   const [totalUSDValue, setTotalUSDValue] = useState('0')
+  // Used when the user explicitly choses proportional input mode
+  const [wantsProportional, setWantsProportional] = useState(false)
```
agualis, line 45: I kept the original wantsProportional naming. Maybe there's a better term but for me it clicks.

packages/lib/modules/pool/actions/add-liquidity/handlers/ProportionalBoostedUnbalancedAddLiquidityV3.handler.ts
```diff
@@ -0,0 +1,40 @@
+import { AddLiquidityBoostedBuildCallInput, AddLiquidityBoostedV3, Hex } from '@balancer/sdk'
+import { TransactionConfig } from '@repo/lib/modules/web3/contracts/contract.types'
+import { SdkBuildAddLiquidityInput } from '../add-liquidity.types'
+import { BaseProportionalAddLiquidityHandler } from './BaseProportionalAddLiquidity.handler'
+import { constructBaseBuildCallInput } from './add-liquidity.utils'
+
+export class ProportionalBoostedUnbalancedAddLiquidityV3Handler extends BaseProportionalAddLiquidityHandler {
+  public async buildCallData({
```
agualis, line 8: There are some duplication  in this and the other v3 proportional handler but I preferred to keep it explicit for now.

PR #223, opened 2024-12-02
feat: add liquidity tabs2

agualis:


packages/lib/shared/components/btns/button-group/ButtonGroup.tsx
```diff
@@ -1,66 +1,141 @@
...
-            </Button>
+          return option?.tabTooltipLabel ? (
+            <Box flex="1" key={`button-group-option-${option.value}`}>
+              <Popover trigger="hover">
+                <PopoverTrigger>
+                  <Box _hover={{ opacity: 0.75 }} transition="opacity 0.2s var(--ease-out-cubic)">
+                    <GroupOptionButton isActive={isActive} option={option} {...props} />
+                  </Box>
+                </PopoverTrigger>
+                {option?.tabTooltipLabel && (
```
groninge01, line 58: when the label is set correctly this is not needed

packages/lib/shared/components/btns/button-group/ButtonGroup.tsx
```diff
@@ -1,66 +1,141 @@
...
+                    </Text>
+                  </PopoverContent>
+                )}
+              </Popover>
+            </Box>
+          ) : (
+            <Box flex="1" key={`button-group-option-${option.value}`}>
+              <GroupOptionButton
+                isActive={isActive}
+                key={`button-group-option-${option.value}`}
```
groninge01, line 71: key is probably not needed here

packages/lib/shared/components/btns/button-group/ButtonGroup.tsx
```diff
@@ -1,66 +1,141 @@
...
+  groupId,
+  hasLargeTextLabel,
+  onChange,
+}: { option: ButtonGroupOption; isActive: boolean } & Props) {
+  return (
+    <Button
+      bg="transparent"
+      id={`button-group-${option.value}`}
+      isDisabled={option.disabled}
+      key={`button-group-option-${option.value}`}
```
groninge01, line 97: i think this key is also probably not needed

packages/lib/shared/components/btns/button-group/ButtonGroup.tsx
```diff
@@ -1,66 +1,141 @@
...
+      <Box fontSize={hasLargeTextLabel ? 'lg' : undefined} position="relative" zIndex="8">
+        {option.label}
+      </Box>
+    </Button>
+  )
+}
+
+function IconPopover({ option }: { option: ButtonGroupOption }) {
+  if (option.tabTooltipLabel) return null
+  if (!option.iconTooltipLabel) return null
```
groninge01, line 126: `if (option.tabTooltipLabel || !option.iconTooltipLabel) return null`

packages/lib/shared/components/btns/button-group/ButtonGroup.tsx
```diff
@@ -1,66 +1,141 @@
...
+  return (
+    <Popover key={option.iconTooltipLabel} placement="top" trigger="hover">
```
groninge01, line 128: i think the key here can be removed

packages/lib/shared/components/btns/button-group/ButtonGroup.tsx
```diff
@@ -1,66 +1,141 @@
...
+      <PopoverTrigger>
+        <Icon as={Info} />
+      </PopoverTrigger>
+      {option?.iconTooltipLabel && (
```
groninge01, line 132: the condition here is not needed i think

packages/lib/shared/components/errors/UnbalancedAddError.tsx
```diff
@@ -1,30 +1,94 @@
...
+
+export function getErrorLabels(
+  isProportionalSupported: boolean,
+  error?: Error | null
+): UnbalancedErrorLabels | undefined {
+  let errorTitle = 'Token amounts error'
+  let errorMessage = 'Unexpected error. Please ask for support'
+  let proportionalLabel = 'Please use proportional mode.'
+
+  if (!error) return
```
groninge01, line 68: move this to the top of the function

packages/lib/shared/components/btns/button-group/ButtonGroup.tsx
agualis: Removed

packages/lib/shared/components/btns/button-group/ButtonGroup.tsx
agualis: If we had multiple options with no `tabTooltipLabel`, we would need that key do differentiate the different Box components, wouldn't we?

packages/lib/shared/components/errors/UnbalancedAddError.tsx
```diff
@@ -1,30 +1,94 @@
...
+  isInvariantRatioPIErrorMessage,
+  isUnbalancedAddError,
+  isUnbalancedAddErrorMessage,
+} from '../../utils/error-filters'
 import { BalAlertLink } from '../alerts/BalAlertLink'
 import { ErrorAlert } from './ErrorAlert'
+import { useAddLiquidity } from '@repo/lib/modules/pool/actions/add-liquidity/AddLiquidityProvider'
 
 type Props = AlertProps & {
   error?: Error | null
```
groninge01, line 16: why is `error` optional?

packages/lib/shared/components/errors/UnbalancedAddError.tsx
```diff
@@ -1,30 +1,94 @@
...
+
+export type UnbalancedErrorLabels = {
+  errorTitle: string
+  errorMessage: string
+  proportionalLabel: string
+}
+
+export function getErrorLabels(
+  isProportionalSupported: boolean,
+  error?: Error | null
```
groninge01, line 62: why is `error` optional?

packages/lib/shared/utils/error-filters.ts
```diff
@@ -57,7 +57,7 @@ export function isUnbalancedAddError(error?: Error | null): boolean {
   return false
 }
 
-export function isUnbalancedAddErrorMessage(error: Error | null): boolean {
+export function isUnbalancedAddErrorMessage(error?: Error | null): boolean {
```
groninge01, line 60: why is `error` optional?

groninge01 (approved):

packages/lib/shared/components/btns/button-group/ButtonGroup.tsx
groninge01: the key is already on the parent

packages/lib/shared/utils/error-filters.ts
agualis: We call it with` isUnbalancedAddErrorMessage(priceImpactQuery.error)` where error can be undefined.

packages/lib/shared/components/btns/button-group/ButtonGroup.tsx
agualis: Removed

packages/lib/shared/components/btns/button-group/ButtonGroup.tsx
agualis: Removed

packages/lib/shared/utils/error-filters.ts
groninge01: looks like it's defined or null everywhere

packages/lib/shared/components/errors/UnbalancedAddError.tsx
agualis: Refactored

PR #262, opened 2024-12-09
fix: negative sign in zero swap price impact

agualis:
Refactors and fixes some price impact labelling issues.
Adds unit tests to avoid regressions.

**Before**:

**After**:
Screenshot 2024-12-09 at 09 33 13

### DIFF
--- a/packages/lib/modules/price-impact/price-impact.utils.ts
+++ b/packages/lib/modules/price-impact/price-impact.utils.ts
@@ -1,4 +1,4 @@
-import { bn } from '@repo/lib/shared/utils/numbers'
+import { bn, fNum, isNegative, isZero } from '@repo/lib/shared/utils/numbers'
 import BigNumber from 'bignumber.js'
 import { PriceImpactLevel } from './PriceImpactProvider'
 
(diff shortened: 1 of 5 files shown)


packages/lib/modules/price-impact/price-impact.utils.ts
```diff
@@ -87,3 +87,29 @@ export const getPriceImpactExceedsLabel = (priceImpactLevel: PriceImpactLevel) =
...
   }
 }
+
+export function getPriceImpactLabel(priceImpact: string | number | null | undefined) {
+  return priceImpact
+    ? isZero(priceImpact)
+      ? ' (0.00%)'
+      : ` (-${fNum('priceImpact', priceImpact)})`
+    : ''
+}
```
alter-eggo, line 97: difficult to read such ternary conditions.. maybe smth like this?
```
export function getPriceImpactLabel(priceImpact: string | number | null | undefined): string {
  if (!priceImpact) {
    return '';
  }

  if (isZero(priceImpact)) {
    return ' (0.00%)';
  }

  return ` (-${fNum('priceImpact', priceImpact)})`;
}
```

alter-eggo (approved):

packages/lib/modules/price-impact/price-impact.utils.ts
agualis: Good catch, I simply moved the function and forgot about refactoring. Done.

PR #268, opened 2024-12-09
feat: move vebal lock provider

alter-eggo:
moved to frontend v3 app, since not needed in beets


groninge01 (approved):
this works for beets
so i hope you tested it for balancer :)

PR #273, opened 2024-12-09
fix: isBoosted should only apply to v3 pools

agualis:
is `isBoosted` logic exclusive to v3? Please confirm before merging. 

https://balancerecosystem.slack.com/archives/C02F9EMRKQA/p1733775808481719


mendesfabio (approved):
looking good - verified Boosted labels only exist for v3 and add/remove liquidity is working again for v2 pools.

PR #277, opened 2024-12-10
fix: pool swap when shouldChangeNetwork

agualis:
- Fixes swap router setup. 
- Fixes issue when pool swap network does not match wallet network.


garethfuller (approved):

agualis:
There is still one issue when the sign step hands in "Preparing state". 
@groninge01 I believe this was already an issue and I will fix it in a different PR.

PR #280, opened 2024-12-10
fix: avoid underlying tokens in not v3 pools

agualis:
Fixes token helpers for !v3 pools with `ERC4626`. 

Pool to test:
https://mono-test-v3-git-fix-v2boosted-balancer.vercel.app/pools/gnosis/v2/0xbc2acf5e821c5c9f8667a36bb1131dad26ed64f9000200000000000000000063

### DIFF
--- a/packages/lib/modules/pool/pool.helpers.ts
+++ b/packages/lib/modules/pool/pool.helpers.ts
@@ -333,7 +333,7 @@ export function isV2Pool(pool: Pool): boolean {
   return pool.protocolVersion === 2
 }
 
-export function isV3Pool(pool: Pool | PoolListItem): boolean {
+export function isV3Pool(pool: Pool | PoolListItem | GqlPoolBase): boolean {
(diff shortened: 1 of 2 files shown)


packages/lib/modules/pool/pool.helpers.ts
```diff
@@ -514,12 +517,22 @@ export function allPoolTokens(pool: Pool | GqlPoolBase): TokenCore[] {
   return uniqBy(allTokens, 'address')
 }
 
-// Returns top level standard tokens + Erc4626 underlying tokens
+function shouldUseUnderlyingToken(token: PoolToken, pool: Pool | GqlPoolBase): boolean {
+  if (isV3Pool(pool) && token.isErc4626 && !token.underlyingToken) {
```
agualis, line 521: v2 pools without `underlyingToken` would work but I prefer to be strict for now to detect potential API issues.

garethfuller (approved):

PR #282, opened 2024-12-10
chore: go back to 15 minutes for tags cache

agualis:
Once we finish the tags setup we will recover the 15 min cache for tags.
So it will be merged after this temporary one https://github.com/balancer/frontend-monorepo/pull/281


Xeonus (approved):

garethfuller:
Should we merge this now?

PR #289, opened 2024-12-10
feat: add tooltip for blocked add button

agualis:
Do we prefer an alert with all the reasons?
If we go for the tooltip there are two Add liquidity buttons.

Example: 

Screenshot 2024-12-11 at 13 42 12

### DIFF
--- a/packages/lib/shared/components/poppover/BalPopover.tsx
+++ b/packages/lib/shared/components/poppover/BalPopover.tsx
@@ -1,0 +1,20 @@
+import { Popover, PopoverContent, PopoverTrigger, Text } from '@chakra-ui/react'
+import { PropsWithChildren } from 'react'
+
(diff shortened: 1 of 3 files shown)


packages/lib/shared/components/poppover/BalPopover.tsx
```diff
@@ -0,0 +1,20 @@
+import { Popover, PopoverContent, PopoverTrigger, Text } from '@chakra-ui/react'
+import { PropsWithChildren } from 'react'
+
+type BalPopoverProps = {
+  text: string
+}
+
+export function BalPopover({ children, text }: PropsWithChildren<BalPopoverProps>) {
```
agualis, line 8: Not used yet but we could iterate this new component to clear Popover related code

garethfuller (approved):

PR #296, opened 2024-12-11
fix: improve chainSlug handling

agualis:
Improves safety by enforcing chain slug check. 
Example of wrong url: 
https://mono-test-v3-git-fix-chainslugs-balancer.vercel.app/pools/mainnet/v2/0x3de27efa2f1aa663ae5d458857e731c129069f29000200000000000000000588

Before: 
Screenshot 2024-12-11 at 10 53 10

After: 
Screenshot 2024-12-11 at 10 51 43

TODO: 
- improve API error handling for other cases


packages/lib/modules/pool/PoolDetail/PoolActivity/usePoolActivity.tsx
```diff
@@ -41,7 +41,7 @@ function _usePoolActivity() {
   const [isExpanded, setIsExpanded] = useState(false)
   const { isChartView } = usePoolActivityViewType()
 
-  const _chain = slugToChainMap[chain as ChainSlug]
+  const _chain = getChainSlug(chain as ChainSlug)
```
agualis, line 44: The type of chain here is `string | string[]`
Is that cast safe (not familiar with this code).

PR #311, opened 2024-12-12
fix: disable proportional adds in WeightedPool2Tokens pools

agualis:
Proportional adds were never supported by the SDK for these old v2 pools (`WeightedPool2Tokens`).
No one had complained probably cause no one had tried the proportional mode, which confirms that the new tab makes it more  accesible.

In that PR I hardcoded the address of the WeightedPool2Tokens factory but we need to understand if other v2 pools has the same incompatibility.


packages/lib/modules/pool/actions/add-liquidity/form/AddLiquidityFormTabs.tsx
```diff
@@ -36,7 +37,9 @@ export function AddLiquidityFormTabs({
   const { toCurrency } = useCurrency()
 
   const isDisabledProportionalTab =
-    nestedAddLiquidityEnabled || !supportsProportionalAddLiquidityKind(pool)
+    nestedAddLiquidityEnabled ||
+    !supportsProportionalAddLiquidityKind(pool) ||
+    hasWeightedPool2Tokens(pool)
```
garethfuller, line 42: Looks like `hasWeightedPool2Tokens` should be within `supportsProportionalAddLiquidityKind`.

packages/lib/modules/pool/actions/LiquidityActionHelpers.ts
```diff
@@ -291,6 +291,22 @@ export function supportsProportionalAddLiquidityKind(pool: Pool): boolean {
   return true
 }
 
+// v2 Pools with WeightedPool2Tokens pool types do not support AddLiquidityKind.Proportional in the SDK
+// TODO: get this data from the API
+export function hasWeightedPool2Tokens(pool: Pool): boolean {
```
garethfuller, line 296: Do you know for sure that it's limited to 2 token pools? My understanding is that we had a couple of versions of weighted pool and all could have more than 2 tokens.

agualis: I'm talking to the SDK/SC teams to confirm. I believe we can merge and create another PR when we have that info.

packages/lib/modules/pool/actions/LiquidityActionHelpers.ts
```diff
@@ -288,9 +288,24 @@ export function supportsProportionalAddLiquidityKind(pool: Pool): boolean {
...
+  if (isWeightedPool2Tokens(pool)) return false
   return true
 }
 
+export function isWeightedPool2Tokens(pool: Pool): boolean {
+  if (
+    isV2Pool(pool) &&
+    isSameAddress(
+      (pool?.factory as Address) || '',
+      getNetworkConfig(pool.chain).contracts.balancer.WeightedPool2TokensFactory || '0xUndefined'
```
garethfuller, line 301:
```suggestion
      getNetworkConfig(pool.chain).contracts.balancer?.WeightedPool2TokensFactory || '0xUndefined'
```

garethfuller (changes requested):

garethfuller (approved):

PR #321, opened 2024-12-15
fix: allowance query

agualis:
- Fixes initial selectedChain logic for Pool Swaps. 
- Avoids calling `allowance` `readContracts` query when  the params are not loaded.

### DIFF
--- a/packages/lib/modules/swap/SwapProvider.tsx
+++ b/packages/lib/modules/swap/SwapProvider.tsx
@@ -228,7 +228,6 @@ export function _useSwap({ poolActionableTokens, pool, pathParams }: SwapProvide
   }
 
   function setSelectedChain(_selectedChain: GqlChain) {
-    if (isPoolSwapUrl) return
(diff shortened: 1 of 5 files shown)


packages/lib/modules/swap/SwapProvider.tsx
```diff
@@ -502,7 +501,9 @@ export function _useSwap({ poolActionableTokens, pool, pathParams }: SwapProvide
   // Sets initial swap state for pool swap edge-case
   function setInitialPoolSwapState(pool: Pool) {
     const { tokenIn } = pathParams
-    setInitialChain(pool.chain)
+    const slugChain = chainToSlugMap[pool.chain]
```
agualis, line 504: This was not causing a bug cause we had `if (isPoolSwapUrl) return` in `setSelectedChain` which becomes unnecessary with this fix.

garethfuller (approved):

PR #322, opened 2024-12-15
feat: emergency boolean flag to block all v3 pool adds

agualis:


packages/lib/modules/pool/pool.helpers.ts
```diff
@@ -278,11 +278,16 @@ export function hasHooks(pool: Pool): boolean {
   return !![pool.hook, ...nestedHooks].filter(Boolean).length
 }
 
+// Emergency flag to block adds for all V3 pools
+const shouldBlockV3PoolAdds = false
```
agualis, line 282: @garethfuller We could do this by project (balancer/beets) in the config setup, or adding a `protocolVersion` to allow blocking v2 instead of v3 but I think this is enough.

garethfuller (approved):

PR #326, opened 2024-12-15
fix: upgrade viem to 2.21.55

groninge01:


apps/beets-frontend-v3/package.json
```diff
@@ -34,7 +34,7 @@
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "react-use-measure": "^2.1.1",
-    "viem": "^2.21.18",
+    "viem": "^2.21.55",
```
agualis, line 37: Better remove ^ to fix the versions as we update them (we prefer fix version for security reasons).

@garethfuller tried to fix them all at once and there was a problem so let's do it bit by bit.

agualis (changes requested):

apps/beets-frontend-v3/package.json
groninge01: fixed the versions

garethfuller (approved):

PR #327, opened 2024-12-16
chore: Minor fixes

garethfuller:
If you try to input a number before a token is selected you get this error. We can fix this by disabling the input until a token is selected. I'm not sure if this will cause issues elsewhere. It shouldn't.
Screenshot 2024-12-16 at 10 25 38

We can hide this input gradient when no token is selected to avoid this weird shadow issue.
Screenshot 2024-12-16 at 10 25 30

### DIFF
--- a/packages/lib/modules/tokens/TokenInput/TokenInput.tsx
+++ b/packages/lib/modules/tokens/TokenInput/TokenInput.tsx
@@ -262,6 +262,7 @@ export const TokenInput = forwardRef(
                 boxShadow="none"
                 fontSize="3xl"
                 fontWeight="medium"
+                isDisabled={!token}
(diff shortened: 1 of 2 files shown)


packages/lib/modules/tokens/TokenInput/TokenInput.tsx
```diff
@@ -262,6 +262,7 @@ export const TokenInput = forwardRef(
                 boxShadow="none"
                 fontSize="3xl"
                 fontWeight="medium"
+                isDisabled={!token}
```
agualis, line 265: This could have an impact in my Api Tokens refactor PR. I will take a look when I can.

PR #335, opened 2024-12-17
Experimental turbopack dev mode

agualis:
Dependency updates and minimum fixes to make `turbopack` work. 

Just run `pnpm run dev:turbopack` (or `pnpm run dev:beets:turbopack`)   and enjoy the new DX 🚀

Added scripts and readme to use this option as an experimental feature with some issues: 
- There is a hydration error.
- Sentry does not in turbopack mode but will keep on working as always in dev and build, webpack based,  modes.
- There are deprecation warnings in style foundations. They are only happening in turbopack mode (not in webpack mode) so we can ignore them for now)

Note that `turbopack` is not ready for builds so we need to keep the current `webpack` config for builds.


packages/lib/modules/pool/PoolDetail/PoolActivityTable/PoolActivityTableRow.tsx
```diff
@@ -58,7 +58,7 @@ function EnsOrAddress({ userAddress, chain }: { userAddress: `0x${string}`; chai
           backgroundColor="background.level4"
           borderRadius="100%"
           height="24px"
-          src={ensAvatar || fallbackSVG.toDataUriSync()}
+          src={ensAvatar || fallbackSVG.toDataUri()}
```
garethfuller, line 61: Can't do this because `toDataUri` returns a promise. That's why Sync was required. Code will have to be refactored if for some reason we've lost `toDataUriSync`.

garethfuller (commented):
They hydration error is an issue in Chakra itself. This PR is blocked unless we can upgrade chakra to fix it.
Screenshot 2024-12-20 at 09 55 12

agualis:
> They hydration error is an issue in Chakra itself. This PR is blocked unless we can upgrade chakra to fix it. 
If that's the only issue I don't see it as a big blocker:
We can create a new script in `package.json`:
` "dev:turbo": "next dev --turbo",`
so that developers can chose this optional "experimental" mode (knowing that it has some hydration issues).

The remaining real blocker is: 
> Can't` do this because toDataUri returns a promise. That's why Sync was required. Code will have to be refactored if for some reason we've lost toDataUriSync.
It comes from `dicebear` update (required to make it compatible with turbopack) and we should be able to fix it to make it work as it used to.

That said, in the medium term I would try to migrate to chakra 3 (only if its compat with nextjs 14/15 is production ready).

packages/lib/modules/pool/PoolDetail/PoolActivityTable/PoolActivityTableRow.tsx
agualis: It used to return a promise but now `toDataUri` is synchronous since version: 
https://github.com/dicebear/dicebear/releases/tag/v9.0.0

apps/beets-frontend-v3/app/layout.tsx
```diff
@@ -23,11 +21,9 @@ export default function RootLayout({ children }: PropsWithChildren) {
         style={{ marginRight: '0px !important' }} // Required to prevent layout shift introduced by Rainbowkit
         suppressHydrationWarning
       >
-        <ColorThemeProvider defaultTheme={DEFAULT_THEME_COLOR_MODE}>
-          <ThemeProvider>
-            <Providers>{children}</Providers>
-          </ThemeProvider>
-        </ColorThemeProvider>
+        <ThemeProvider>
```
garethfuller, line 24: Why remove ColorThemeProvider here?

packages/lib/modules/tokens/TokenIcon.tsx
```diff
@@ -76,7 +76,7 @@ export function TokenIcon({
           borderRadius="100%"
           height={`${size}px`}
           onError={() => !hasError && setHasError(true)}
-          src={hasError || !iconSrc ? fallbackSVG.toDataUriSync() : iconSrc}
+          src={hasError || !iconSrc ? fallbackSVG.toDataUri() : iconSrc}
```
garethfuller, line 79: This doesn't work. `toDataUri` returns a promise, we need to use `toDataUriSync` or find another way to do this.

packages/lib/shared/components/btns/DarkModeToggle.tsx
```diff
@@ -47,7 +36,7 @@ export default function DarkModeToggle() {
   return (
     <Button onClick={toggleColorMode} p="0" variant="tertiary">
       <AnimatePresence initial={false}>
-        {theme === 'light' ? (
+        {colorMode === 'dark' ? (
```
garethfuller, line 39: Why has this changed?

apps/beets-frontend-v3/app/layout.tsx
agualis: Groninge removed it from beets to demonstrate that the hydration issue is due to this (beets only has dark mode)

packages/lib/modules/tokens/TokenIcon.tsx
agualis: Did you run pnpm install? Did you see this comment? 
https://github.com/balancer/frontend-monorepo/pull/335#issuecomment-2556846355

packages/lib/shared/components/btns/DarkModeToggle.tsx
groninge01: looking at other sites this seems to be the consensus: show the icon for the state you can go _to_ instead the _current_ state
i also asked @uiuxxx about it

apps/beets-frontend-v3/app/layout.tsx
groninge01: it looks like next-themes is no longer needed but i'll admit i only looked briefly
as this only concerns balancer it might need more testing

garethfuller: @agualis can you look into theming with chakra/nextjs. It used to be that using next-themes was the default approach. I think there might have also been a reason to use it so that the theme applied correctly in RSCs.

packages/lib/shared/components/btns/DarkModeToggle.tsx
uiuxxx: I don't think there is really UX consensus around this because there are a couple of conflicting rules and both options are a little confusing.

UIs ideally should communicate current status, while buttons should indicate the action to occur (not the current state). Also, some sites use toggles which has the current state highlighted. 

So for cleanliness a lot of sites combine the button with the current state (e.g. https://tailwindcss.com/). But yeah, it breaks the rule of buttons indicating actions...

If I had to choose, I'd do it per @groninge01 – have the button icon denote the state where it is going.  

But yeah, perhaps we should introduce a toggle in the future to make this more obvious.

garethfuller (approved):

PR #365, opened 2024-12-19
fix: Token inputs

garethfuller:
Before:
Screenshot 2024-12-19 at 15 25 55

After:
Screenshot 2024-12-19 at 16 02 48

### DIFF
--- a/packages/lib/modules/pool/actions/add-liquidity/form/TokenInputs.tsx
+++ b/packages/lib/modules/pool/actions/add-liquidity/form/TokenInputs.tsx
@@ -31,9 +31,8 @@ export function TokenInputs({ tokenSelectDisclosureOpen, customSetAmountIn }: Pr
 
   function weightFor(tokenAddress: string) {
     return (
-      pool.poolTokens.find(token =>
-        isSameAddress(token.address as Address, tokenAddress as Address)
-      )?.weight ?? undefined
+      tokens.find(token => isSameAddress(token.address as Address, tokenAddress as Address))
+        ?.weight ?? undefined
     )
   }
 
@@ -44,6 +43,7 @@ export function TokenInputs({ tokenSelectDisclosureOpen, customSetAmountIn }: Pr
 
         return (
           <TokenInput
+            address={token.address}
             apiToken={token}
             chain={token.chain}
             isDisabled={hasNoLiquidity(pool)}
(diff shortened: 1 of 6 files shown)


packages/lib/modules/pool/actions/add-liquidity/form/TokenInputs.tsx
```diff
@@ -31,9 +31,8 @@ export function TokenInputs({ tokenSelectDisclosureOpen, customSetAmountIn }: Pr
 
   function weightFor(tokenAddress: string) {
     return (
-      pool.poolTokens.find(token =>
-        isSameAddress(token.address as Address, tokenAddress as Address)
-      )?.weight ?? undefined
```
garethfuller, line 36: This doesn't work because if the tokenAddress in question is an underlying token it doesn't find the weight.

packages/lib/modules/pool/actions/add-liquidity/form/TokenInputs.tsx
```diff
@@ -31,9 +31,8 @@ export function TokenInputs({ tokenSelectDisclosureOpen, customSetAmountIn }: Pr
...
+      tokens.find(token => isSameAddress(token.address as Address, tokenAddress as Address))
+        ?.weight ?? undefined
```
garethfuller, line 35: This works because in the `tokens` construction the high level token and it's underlying are merged so that the weight is attribute is preserved. `{ ...token, ...underlyingToken }`

packages/lib/modules/pool/pool.helpers.ts
```diff
@@ -615,7 +615,7 @@ export function getBoostedGqlTokens(pool: Pool): ApiToken[] {
   const underlyingTokens = pool.poolTokens
     .flatMap(token =>
       shouldUseUnderlyingToken(token, pool)
-        ? [token.underlyingToken as ApiToken]
+        ? [{ ...token, ...token.underlyingToken } as ApiToken]
```
garethfuller, line 618: Need to merge in attributes from high level token so we can pass on the weight attribute.

packages/lib/modules/pool/actions/add-liquidity/form/TokenInputs.tsx
garethfuller: It could break for nested pools?

agualis (approved):

PR #373, opened 2024-12-20
refactor: PoolCore, remove pool.displayTokens, new display token functions

agualis:
This goes on top of the big [voting list PR](https://github.com/balancer/frontend-monorepo/pull/148) because that contains many displayToken related changes that would conflict the the planned tokens refactor.

Main changes: 

- Removes all API deprecated `pool.displayTokens` calls.
- Simplifies token related components types by using `ApiToken` instead of types like `GqlDisplayToken` 
- Centralises token display helpers in a new file (`pool.tokens.display.ts`). Note that this is just a temporary file. Its methods and the file itself will be renamed/moved in another PR (where we will fix the logic related with pool actions and `getPoolActionableTokens`).
- Sorts display pool tokens by index to have a more predictable token display outputs. 
- Adds integration tests to cover all the new extracted functions above (covering all the pool edge cases).
- Introduces a new `PoolCore` type to start cleaning the mess around different Pool types (`Pool, PoolListItem, PoolData`...) used across the Pool token display related components. (This must be simplified even more in incoming PRs but we need it now to enable previous points).

`TODO`: 
The refactor is far from perfect but should unblock Voting list PR and simplify the path to a cleaner code base.

- [x] Cyclic dependency in `pool.tokens.display`: Fixed in https://github.com/balancer/frontend-monorepo/pull/378
- [x] TokenRow the pool token related logic is messy, we need to find proper abstractions and functions for `pool.tokens.utils.ts`
- [x]  `VoteListTableRow`: Different problems with pool structure (check TODOs in that file)
- [x] Rename long functions like `getCompositionDisplayTokensWithPossibleNestedTokensWithBalance`
- [ ]  Rethink all the logic around PoolCore, Pool, FeaturedPool, PoolListItem, GqlPoolElement...
- [ ] Do we need balance in underlying tokens? (review and test all posible edge cases)
- [ ] There are 2 `TokenRow` components (one in `TokenRowSelectList`)with optional props like pool, token, apiToken...Review if we can simplify.
- [ ] Extract a Fragment for FeaturedPools and create a PR to remove them (they are not currently used)


apps/beets-frontend-v3/app/(app)/pools/[chain]/[variant]/[id]/layout.tsx
```diff
@@ -48,7 +44,7 @@ export async function generateMetadata({
   const pool = data?.pool
   if (!pool) return {}
 
-  const displayTokens = getPoolDisplayTokens(pool)
+  const displayTokens = getHeaderDisplayTokens(pool as PoolCore)
```
Comment (garethfuller) [inline apps/beets-frontend-v3/app/(app)/pools/[chain]/[variant]/[id]/layout.tsx:47]: Why is this any different to getActionTokens?

packages/lib/modules/pool/PoolList/PoolListTokenPills.tsx
```diff
@@ -172,47 +165,36 @@ function StableTokenPills({
...
 
   // TODO: Move this into a general 'displayTokens' helper function.
   if (isV3Pool(pool) && pool.hasErc4626 && pool.hasAnyAllowedBuffer) {
     poolTokens = poolTokens.map(token =>
       token.isErc4626 && token.isBufferAllowed
-        ? ({ ...token, ...token.underlyingToken } as unknown as GqlPoolTokenDetail)
+        ? ({ ...token, ...token.underlyingToken } as unknown as ApiToken)
         : token
     )
   }
```
garethfuller, line 194: What is the point of keeping this logic if `poolTokens` isn't used?

packages/lib/modules/pool/pool.tokens.display.ts
```diff
@@ -0,0 +1,108 @@
...
+
+// CRITICAL TODO: duplicated function as importing from pool.helpers creates a circular dependency or something
+// (use the import from pool.helpers to see the error)
+export function isV3Pool(
+  pool: Pick<Pool | PoolListItem | GqlPoolBase, 'protocolVersion'>
+): boolean {
+  return pool.protocolVersion === 3
+}
+
+export function getCompositionDisplayTokens(pool: PoolCore | GqlNestedPool): ApiTokenWithBalance[] {
```
garethfuller, line 18: I don't think we need to keep using the word Display. It's kind of irrelevant if we are classifying them as Composition/Header. The fact that they are displayed doesn't mean anything. displayTokens is a legacy thing where 'display' was the classification.

```suggestion
export function getCompositionTokens(pool: PoolCore | GqlNestedPool): ApiTokenWithBalance[] {
```

packages/lib/modules/pool/pool.tokens.display.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+    return token as ApiTokenWithBalance
+  })
+
+  return sortApiTokensBySymbol(
+    excludeNestedBptTokens(tokens as ApiToken[], pool.address)
+  ) as ApiTokenWithBalance[]
+}
+
+export function getHeaderDisplayTokens(pool: PoolCore): ApiToken[] {
```
garethfuller, line 34: I'm not sure Header is the right word because it's for the pool header and list. Trying to describe them here: They are reference tokens for the user so that they know at a glance what can be added to the pool. maybe?

```suggestion
// The set of tokens that are reference tokens for the user so that they know at a glance what can be added to the pool. Used in the pool header and in the pools list.
export function getUserReferenceTokens(pool: PoolCore): ApiToken[] {
```

packages/lib/modules/pool/pool.tokens.display.ts
```diff
@@ -0,0 +1,108 @@
...
+  return sortBy(tokens, 'symbol')
+}
+
+function excludeNestedBptTokens(tokens: PoolToken[] | ApiToken[], poolAddress: string): ApiToken[] {
+  return tokens
+    .filter(token => !isSameAddress(token.address, poolAddress as Address)) // Exclude the BPT pool token itself
+    .filter(token => token !== undefined) as ApiToken[]
+}
+
+export function getHeaderTokensWithPossibleNestedTokensWithBalance(
```
garethfuller, line 77: I really dislike long function names like this. Should these not be flags to the underlying function?
```
getHeaderTokens(pool, { includeNested: true, includeBalance: true }).
```

packages/lib/modules/pool/PoolDetail/PoolComposition.tsx
```diff
@@ -181,7 +185,9 @@ export function PoolComposition() {
           ) : (
             <PoolWeightChart
               chain={chain}
-              displayTokens={getPoolDisplayTokensWithPossibleNestedPools(pool)}
+              displayTokens={getCompositionDisplayTokensWithPossibleNestedTokensWithBalance(
```
garethfuller, line 188: This is a big code smell.

garethfuller (changes requested):

packages/lib/modules/pool/pool.tokens.display.ts
garethfuller, line 1: Rename `pool-tokens.utils.ts`

agualis:
Replaced by https://github.com/balancer/frontend-monorepo/pull/427 to simplify review and merge process.

PR #376, opened 2024-12-20
merge from beets

groninge01:


packages/lib/modules/web3/contracts/useMulticall.ts
```diff
@@ -33,6 +36,9 @@ export function useMulticall(multicallRequests: ChainContractConfig[], options:
           const results = await multicall(config, {
             contracts: multicalls,
             chainId: Number(chain),
+            ...(Number(chain) === getChainId(GqlChain.Sonic) && {
```
agualis, line 39: Created a task to not forget about removing this: 
https://github.com/orgs/balancer/projects/12/views/1?pane=issue&itemId=92621002&issue=balancer%7Cfrontend-monorepo%7C399

agualis (approved):
As a general comment, we could improve the config around code using `isBalancerProject` but I think you already talked about this with @garethfuller 

Approved but let's wait for Gareth to do another review.

apps/frontend-v3/lib/components/navs/NavBarContainer.tsx
```diff
@@ -22,10 +25,16 @@ export function NavBarContainer() {
       >
         <NavBar
           appLinks={allAppLinks}
+          customLinksAfter={
```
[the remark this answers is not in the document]
groninge01: changed it!

packages/lib/modules/transactions/transaction-steps/receipts/receipt-parsers.ts
```diff
@@ -149,6 +152,39 @@ export function parseSwapReceipt({
...
+export function parseLstWithdrawReceipt({ receiptLogs, userAddress, chain, getToken }: ParseProps) {
+  const receivedToken: HumanTokenAmount[] = getIncomingLogs(receiptLogs, userAddress).map(log => {
+    const tokenDecimals = getToken(log.address, chain)?.decimals
+    return _toHumanAmount(log.address, log.args.value, tokenDecimals)
+  })
+
+  return {
+    receivedToken,
+  }
+}
```
garethfuller, line 186: You shouldn't need to parse receipts for staking/unstaking because there is no
slippage. The only reason we do receipt states in the add/remove/swap flow is
because the amounts can change between preview and confirmation. I think it's
likely all this can be removed.

garethfuller (changes requested):
In general looks good, my biggest concern is the receipt parsers/steps for LST staking. I'm pretty sure we shouldn't be doing that. Like we don't use receipt steps for staking/unstaking bpt. Could be wrong though?

apps/frontend-v3/lib/components/navs/NavBarContainer.tsx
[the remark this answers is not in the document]
garethfuller: Thanks, sorry didn't see the full component change. It's not a big deal, we can leave it until we need a links before.

packages/lib/modules/transactions/transaction-steps/receipts/receipt-parsers.ts
groninge01: for the unstaking you are right, there is no token transferred at that stage so i have removed it

for staking and withdrawing though, tokens are transferred and i have updated the parsers to correctly parse the logs for that
i also updated the tests for there 2 parsers

PR #414, opened 2025-01-07
fix: transaction action button

alter-eggo:
bug [context](https://balancerecosystem.slack.com/archives/C02F9EMRKQA/p1736221312345609)

we need to provide key to create new RenderActionButton's instance depending on currentStep.id otherwise it may take previous transaction result from `useManagedTransaction`


agualis (approved):
@anastas 
Approved but I don't really understand how this can happen. Maybe related with the `motion.div` wrapper?

garethfuller (approved):

PR #423, opened 2025-01-09
feat: add middle ware for cors

groninge01:


groninge01:
looks like cors won't matter because the request will always come through beets.fi

PR #440, opened 2025-01-15
update: lst unstake & withdrar summaries

groninge01:


apps/beets-frontend-v3/lib/modules/lst/components/LstUnstakeSummary.tsx
```diff
@@ -25,21 +19,12 @@ export function LstUnstakeSummary() {
       <Card variant="modalSubSection">
         <LstTokenRow
           chain={chain}
-          isLoading={isLoading}
+          isLoading={false}
```
agualis, line 22: Tiny improvement: default prop to false so that it is not required.

agualis (approved):
Approved with small suggestion.

apps/beets-frontend-v3/lib/modules/lst/components/LstUnstakeSummary.tsx
groninge01: done!

agualis (approved):

PR #463, opened 2025-01-17
chore: remove unused Global Alert provider

agualis:


garethfuller (commented):
We may at some point need global alerts though?

agualis:
> We may at some point need global alerts though?
@garethfuller The current implementation was not working as expected and I wanted to avoid global providers on top of `Web3Provider` to discard that affecting disconnection errors (like the one you fixed in https://github.com/balancer/frontend-monorepo/pull/462)

If we eventually need `GlobalProviders` we can take a look at the deleted code.

garethfuller (approved):

PR #469, opened 2025-01-20
fix: initial networkConfig

agualis:
Copy pasting a swap url with token in and out was not setting the initial tokens (in and out) cause it was using `mainnet` `networkConfig` by default. 

In this PR we enforce that the `networkConfig` is the one that we load in `setInitialChain`

Loading `swap/polygon/0x2c89bbc92bd86f8075d1decc58c7f4e0107f286b/WETH`

**Before**: 
Screenshot 2025-01-20 at 12 37 49

**After**:
after


garethfuller (approved):

garethfuller:
Will clash with @groninge01 config PR.

agualis:
> Will clash with @groninge01 config PR.

I will wait for his merge to solve the conflicts.

PR #485, opened 2025-01-22
feat: update block add liquidity w/ erc4626 rules

### LINKED ISSUE #318: Block adds for erc4626 reviews
Block add liquidity depending on erc4626 review data. Should be the same as rate provider logic. See `shouldBlockAddLiquidity`

groninge01:
fixes #318


garethfuller (commented):
There seems to be an e2e test failure here in the beets app?
Screenshot 2025-01-22 at 11 01 05

groninge01:
> There seems to be an e2e test failure here in the beets app? Screenshot 2025-01-22 at 11 01 05

good test
i need to check to see if a token is an erc4626 before checking the rest...

garethfuller (approved):

PR #486, opened 2025-01-22
feat: my votes table and submit

alter-eggo:
This pr contains:
- my votes table
- submit logic

pr seems big, though most of changes are quite simple and there is a lot of layout work


packages/lib/modules/vebal/vote/VoteList/VoteListLayout.tsx
```diff
@@ -40,6 +45,26 @@ export function VoteListLayout() {
...
+            description: selectedVotingPool.tokens
+              .map(token => `${token.symbol}-${token.weight ?? 0}`)
+              .join('-'),
+          }))}
+        />
+      )
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedVotingPools])
+
```
garethfuller, line 67: I don't get this pattern? Did you pick it up from documentation somewhere so I can learn? Why not just have the <SelectedPoolsMenu /> component within the toast component?

packages/lib/modules/vebal/vote/VoteList/VoteListTable/VoteListTable.tsx
```diff
@@ -23,12 +24,28 @@ export function VoteListTable({ voteList, count, loading }: Props) {
...
-  }
+  const rowProps = useMemo(
+    () => ({
+      px: { base: 'sm', sm: '0' },
+      gridTemplateColumns: `32px minmax(320px, 1fr) 120px 100px 120px 120px 100px`,
+      alignItems: 'center',
+      gap: { base: 'xxs', xl: 'lg' },
+    }),
+    []
+  )
```
garethfuller, line 35: This is a static object... It does not need to be in a useMemo.

packages/lib/modules/vebal/vote/Votes/MyVotes/MyVotesTable/MyVotesTable.tsx
```diff
@@ -0,0 +1,121 @@
...
+    () => ({
+      px: { base: 'sm', sm: '0' },
+      gridTemplateColumns: `32px minmax(320px, 1fr) 100px 120px 120px 120px 50px`,
+      alignItems: 'center',
+      gap: { base: 'xxs', xl: 'lg' },
+      // fix: (votes) implement nested cell paddings for Edit votes column
+      // gridTemplateColumns: `52px minmax(350px, 1fr) 130px 150px 150px 150px 52px`,
+    }),
+    []
+  )
```
garethfuller, line 37: Again, why is this a memo?

packages/lib/modules/vebal/vote/Votes/MyVotes/MyVotesTable/MyVotesTable.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  const cellProps = useMemo(
+    () => ({
+      // fix: (votes) implement nested cell paddings for Edit votes column
+      // p: ['sm', 'md'],
+    }),
+    []
+  )
```
garethfuller, line 45: Shouldn't be a memo.

packages/lib/modules/vebal/vote/Votes/MyVotes/myVotes.types.ts
```diff
@@ -0,0 +1,33 @@
...
+    title:
+      'This shows the ratio of 3rd party voting incentives (known as Bribes) to veBAL. The higher this ratio, the more profitable it is to currently vote on this pool. Note this ratio could change up till the voting deadline.',
+  },
+  currentVotes: {
+    label: 'Current votes',
+  },
+  editVotes: {
+    label: 'Edit votes',
+  },
+}
```
garethfuller, line 33: This is not a type, should be in utils/helpers.

garethfuller (changes requested):

apps/beets-frontend-v3/lib/modules/lst/components/LstWithdraw.tsx
```diff
@@ -55,10 +55,10 @@ export function LstWithdraw({
         noItemsFoundLabel="No requests found"
         paginationProps={{ ...paginationProps, mt: 'auto' }}
         renderTableHeader={() => <LstWithdrawTableHeader {...rowProps} />}
-        renderTableRow={(withdrawal: any, index) => (
+        renderTableRow={({ item, index }) => (
```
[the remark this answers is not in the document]
alter-eggo: I've changed paginated table, so it allows to provide react component instead of common functions, otherwise React will recreate component instance on every render.

Example usage:
// before
renderTableRow={(item, index) => <Component item={item} />})

// after
renderTableRow={Component}

packages/lib/modules/transactions/transaction-steps/TransactionStateProvider.tsx
```diff
@@ -29,9 +29,12 @@ export function _useTransactionState() {
...
 
-  function getTransaction(id: string) {
-    return transactionMap.get(id)
-  }
+  const getTransaction = useCallback(
+    (id: string) => {
+      return transactionMap.get(id)
+    },
+    [transactionMap]
+  )
```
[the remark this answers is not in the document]
alter-eggo: this fn is used in useMemo in submit tx steps, if not put in useCallback => on each render useMemo will recalc

packages/lib/modules/vebal/vote/VoteList/VoteListLayout.tsx
alter-eggo: `StaticToast` depends on children via `useEffect`
```tsx
useEffect(() => {
  if (toastIdRef.current) {
    // rerender
    toast.update(toastIdRef.current, { ...defaultProps, ...props, render: children })
  }
  // eslint-disable-next-line react-hooks/exhaustive-deps
}, [children])
```
 
If we do it this way:
```tsx
<StaticToast isOpen={selectedVotingPools.length > 0}>
  () => <SelectedPoolsMenuRender selectedVotingPools={selectedVotingPools} />
</StaticToast>
```
This will lead to `toast.update` being called on every rerender of `VoteListLayout`. 
With current way, we have `toast.update` happening only when the reference to selectedVotingPools changes.
 
In fact, an example of "ideal" memoization is when you memoize not only props but also children.
[Link to the gist](https://gist.github.com/slikts/e224b924612d53c1b61f359cfb962c06#theres-no-completely-nice-solution)
In most cases, this is excessive, but here it is definitely needed due to `useEffect`.
And I can't remove the `useEffect` from `StaticToast` since I need a way to rerender the created `toast`.
 
Perhaps you want it to be like this:
```tsx
function SelectedPoolsMenuRender(props: ToastProps) {
  // ...
}
 
function VoteListLayout() {
  // ...
  return <StaticToast isOpen={selectedVotingPools.length > 0}>
    {SelectedPoolsMenuRender}
  </StaticToast>
}
```
 
But this won’t work, as we wouldn't be able to pass `selectedVotingPools`. We need a closure magic:
```tsx
const SelectedPoolsMenuRender = useMemo(() => {
  return (props: ToastProps) => <SelectedPoolsMenuRender selectedVotingPools={selectedVotingPools} />
}, [selectedVotingPools])
```
 
And to see generated component's name in React DevTools, we should provide named function:
```tsx
const SelectedPoolsMenuRender = useMemo(() => {
  return function SelectedPoolsRender(props: ToastProps) {
    return <SelectedPoolsMenu selectedVotingPools={selectedVotingPools} />
  }
}, [selectedVotingPools])
```

garethfuller: Ok, but why not this:
```
<StaticToast isOpen={selectedVotingPools.length > 0}>
  <SelectedPoolsMenu selectedVotingPools={selectedVotingPools} />
</StaticToast>
```
Are the rerenders that big on an issue in this case?

alter-eggo: Children is function, so it would be
```
<StaticToast isOpen={selectedVotingPools.length > 0}>
  () => <SelectedPoolsMenuRender selectedVotingPools={selectedVotingPools} />
</StaticToast>
```
About rerenders: I can't predict how toast.update will work in future (maybe it will update some state and we will get infinite rerenders loop), so I try to implement it as safe as possible

garethfuller: Ok, I understand the issue a bit better now and we came both think a refactor is required here. #505

garethfuller (approved):

PR #506, opened 2025-01-24
[WIP] add reliquary to beets

groninge01:


packages/lib/modules/pool/actions/add-liquidity/AddLiquidityProvider.tsx
```diff
@@ -41,7 +41,7 @@ import { useWrapUnderlying } from '../useWrapUnderlying'
 export type UseAddLiquidityResponse = ReturnType<typeof _useAddLiquidity>
 export const AddLiquidityContext = createContext<UseAddLiquidityResponse | null>(null)
 
-export function _useAddLiquidity(urlTxHash?: Hash) {
+export function _useAddLiquidity(urlTxHash?: Hash, relicId?: string) {
```
agualis, line 44: Why do we need this relicId inside the provider if we are not doing anything with it? I would prefer to have related relic logic completely decoupled.

packages/lib/modules/pool/actions/add-liquidity/queries/useAddLiquidityBuildCallDataQuery.ts
```diff
@@ -54,6 +54,8 @@ export function useAddLiquidityBuildCallDataQuery({
     relicId,
   }
 
+  console.log({ relicId })
```
agualis, line 57: Remove

groninge01: Removed!

packages/lib/shared/services/batch-relayer/batch-relayer.service.ts
```diff
@@ -3,18 +3,37 @@ import {
...
+    private readonly reliquaryActionsService?: ReliquaryActionsService
   ) {}
 
-  public encodePeekChainedReferenceValue(reference: bigint): string {
+  static create(batchRelayerAddress: string, includeReliquary = false): BatchRelayerService {
+    return new BatchRelayerService(
+      batchRelayerAddress,
+      gaugeActionsService,
+      vaultActionsService,
+      includeReliquary ?? false ? reliquaryActionsService : undefined
```
agualis, line 32:
```suggestion
      includeReliquary ? reliquaryActionsService : undefined
```
`includeReliquary` is false be default so we don't need `?? false`

groninge01: fixed!

agualis (approved):

Issue #34, opened 2024-06-07
Create wallet


garethfuller:
https://x.com/jessepollak/status/1798926928570966065?s=46

garethfuller:
We already technically have this just by supporting Coinbase wallet in the modal.

garethfuller:
#507

PR #507, opened 2025-01-24
feat: Add create wallet button

garethfuller:
Adds option to create Coinbase wallet. Implements #34 

Screenshot 2025-01-24 at 16 16 44


garethfuller:
@groninge01 is this something we should hide for Beets?

groninge01 (approved):

PR #520, opened 2025-01-28
feat: enable batch transactions for adds in Safe App

agualis:
Example with a v3 pool add (used [this sepolia pool](https://test.balancer.fi/pools/sepolia/v3/0x5b17fb19b8c44f126e87882e7baa32153edcaf1d))

### 1.  Sign permit2 the bundle
If there are pending permit2 token approvals, they are bundled together with the add tx: 
Safev3Bundle

### 2. Safe confirmation
The Safe must confirm the bundle of tx

Safe Confirmation

### 3. Bundle tx success
Transaction bundle status is controlled by the Safee and we finally see the success in the app:

BundleTransactionSucces


garethfuller (commented):
Could we maybe recruit QA to help us do some testing on this?

PR #530, opened 2025-01-30
fix: v3 Safe removes approve BPT with approval tx instead of permit signature

agualis:
Signing standard permit approvals is not possible with Smart Accounts (including Safe):

>  Regular permits (EIP-2612) cannot be directly signed by Gnosis Safe contracts - they can only be signed by EOAs. This is because EIP-2612 permits specifically require ECDSA signatures (specifically eth_sign/personal_sign type signatures), while Gnosis Safe uses contract-based signatures.
> The main reasons:
> Regular permits expect r, s, v components from ECDSA signatures
> Gnosis Safe uses a different signing scheme (contract-based signatures via EIP-1271)
> EIP-2612's permit function validates signatures using ecrecover
> Permit2 works with Gnosis Safe because it was specifically designed to support both EOA signatures and EIP-1271 contract signatures. When using Permit2, the contract checks if the signer is a contract and if so, uses isValidSignature(hash, signature) for verification instead of ecrecover.
> This is one of the major advantages of Permit2 - it’s more flexible in terms of signature validation and works with both EOAs and smart contract wallets

So, in this specific context (Safe App or Safe via WC) we show an `BPT token approval step` instead.

In the future, as an improvement, we could batch the BPT token approval tx + the remove tx.

### DIFF
--- a/packages/lib/modules/pool/actions/remove-liquidity/RemoveLiquidityProvider.spec.tsx
+++ b/packages/lib/modules/pool/actions/remove-liquidity/RemoveLiquidityProvider.spec.tsx
@@ -24,6 +24,7 @@ const simulationQueryResult = {
       aTokenAmountMock(balAddress, balTokenOutUnits),
       aTokenAmountMock(wETHAddress, wEthTokenOutUnits),
     ],
+    sdkQueryOutput: { bptIn: { amount: 100000n } },
   },
 }
 // Mock query to avoid onchain SDK call from unit tests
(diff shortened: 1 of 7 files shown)


garethfuller (commented):
Sorry, didn't manage review before merge. Looks good to me though. I guess the additional functions in useRemoveLiquidityStesp could be moved elsewhere in a refactor. We typically just have a single purpose function in those steps files afaik.

PR #532, opened 2025-01-30
feat: build revert tenderly simulations

agualis:
Add `tenderlyUrl` in sentry meta for unknown revert errors including `Raw Call Arguments`.


packages/lib/shared/utils/errors.ts
```diff
@@ -144,8 +155,21 @@ function parseRequestError(error: Error, chainId: number): TransactionConfig | u
...
+  const dataMatch = error.message.match(/data:\s*([^\n]+)/)
+
+  const to = toMatch?.[1].trim() ?? ''
+  const data = dataMatch?.[1].trim() ?? ''
+
+  return {
+    data: data as Hex,
+    to: to as Address,
+    chainId,
+    account: '0x0000000000000000000000000000000000000000',
```
garethfuller, line 171: Is this not a bit misleading? Should it not be in a specific utils file for tests? Otherwise, should the user's address not be parsed and used?

garethfuller (approved):

Issue #53, opened 2024-07-02
Tooltip text for unknown price impact


uiuxxx:
For unknown price impacts in both 'Add liquidity' and 'Swap' flows, replace the existing default price impact tooltip with the following:

```This usually displays the negative price impact of the swap based on the current market prices of the token in vs token out. However, for some reason, the price impact currently can’t be calculated. This may be due to the pricing provider being down or not knowing one of the tokens. Only proceed if you know exactly what you are doing.```

![Image](https://github.com/balancer/frontend-monorepo/assets/149399376/02e7ff3a-8a29-47f2-87d4-7acd22203175)

SomeCodingUser:
Hey, I took this small task and created a PR, hope this is fine

PR #542, opened 2025-02-02
change tooltip text for unknown price impact (#53)

SomeCodingUser:
Hello,

first time contributing here. Please let me know in case this change is not needed anymore or anything is missing.


garethfuller (commented):
This text change doesn't make sense here. It is rendered for all price impact states.

PR #549, opened 2025-02-03
fix: rainbow kit issue when reloading page connected with WC

agualis:
- Avoids double WC connector instance after reloading page. This is a known issue in RainbowKit implementation and we will use a dependency patch until they review our incoming PR in their repo.

- Also adds WC custom icon for balancer/beets: 

Screenshot 2025-02-03 at 18 14 43


garethfuller (changes requested):
Doesn't work for me. I tried swapping 3 times with reloads in between each and all ended up stuck in pending.

Screenshot 2025-02-04 at 09 03 56

FYI I turned my adblocker off on safe and the preview URL.

Issue #563, opened 2025-02-05
Update yield APR tooltips


uiuxxx:
Update the tooltips as follows:

https://www.figma.com/design/xiTeGFg2XxRbcM2t00b7E2/v3-app-2025?node-id=420-14&t=u9njlTxmyMsN3XkM-1

![Image](https://github.com/user-attachments/assets/2f7cb8dc-2c84-4fe4-a167-da39cbc94e27)

If it's possible (and easy), add logic to display whether the pool is core or not in the Pool attributes metadata.

uiuxxx:
@garethfuller I don't suppose there's an easy way to include the "Learn more" links? Otherwise, we probably should include the actual exact fees in the tooltips.

garethfuller:
@uiuxxx Links are not going to work. We already have a popover withing a popover for this tooltip, getting the link click to work within the pools list is more complicated than it's worth IMO.

PR #564, opened 2025-02-05
chore: Update tooltips

garethfuller:
Implements all text changes suggested in the designs linked here #563.

- Does not include links in tooltips. For some reason links are not clickable within these tooltips and if you click on them within the pool list they redirect to the pool page regardless of the href.
- Does not include Core pool attribute in pool info, this is not available in the API yet.


uiuxxx:
@garethfuller I don't suppose there's an easy way to include the "Learn more" links? Otherwise, we probably should include the actual exact fees in the tooltips.

agualis (approved):

PR #572, opened 2025-02-07
chore: Bump pkgs as per dependabot suggestions.

garethfuller:


agualis (approved):

garethfuller:
@groninge01 there was a minor bump to tanstack react query in the beets app here. Shouldn't affect anything, but could you do a quick sense check to make sure?

PR #576, opened 2025-02-07
Pools page updates

uiuxxx:


groninge01:
as the stats & partner cards are project related can they be made optional (in the project configs)? (if so i could add them)

garethfuller:
@groninge01 yes please adjust as needed.

PR #612, opened 2025-02-13
feat: add smoke test for swap page

agualis:


packages/e2e-tests/tests/balancer/bal.swap.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { test, expect } from '@playwright/test'
+
+test('Open swap page and try eth wrap', async ({ page }) => {
+  await page.goto('http://localhost:3000/')
+  await page.getByRole('link', { name: 'Swap', exact: true }).click()
+
+  // Selects Wrapped Ether token out
+  await page.getByRole('button', { name: 'Select token' }).click()
+  await page.getByPlaceholder('Search by name, symbol or').click()
```
groninge01, line 9: focus should be set here on modal open or doesn't that work for playwright?

packages/e2e-tests/tests/balancer/bal.swap.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+  await page.getByPlaceholder('Search by name, symbol or').fill('we')
+  await page.getByText('Wrapped Ether').click()
+
+  // Fills 1 ETH token in
+  await page.getByPlaceholder('0.00').first().click()
```
groninge01, line 14: same as above here

packages/e2e-tests/tests/balancer/bal.swap.spec.ts
agualis: Probably yes. We can improve this things when we have more bandwidth.

PR #617, opened 2025-02-14
chore: explain proportions for add

groninge01:
image

- [ ] add rules for when to show the info box
- [ ] change icon?


groninge01:
close for now

PR #635, opened 2025-02-19
add rwas filter

mendesfabio:
A few weeks ago we discussed over Slack about having a category to highlight RWA pools. I've just done the obvious changes but might've missed something. Metadata PR has been merged https://github.com/balancer/metadata/pull/46

### DIFF
--- a/packages/lib/config/projects/beets.ts
+++ b/packages/lib/config/projects/beets.ts
@@ -20,7 +20,7 @@ export const ProjectConfigBeets: ProjectConfig = {
   },
   options: {
     poolDisplayType: PoolDisplayType.Name,
-    hidePoolTags: ['VE8020'],
+    hidePoolTags: ['RWA', 'VE8020'],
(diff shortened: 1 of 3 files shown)


groninge01:
with regards to the metadata: 
if a third party creates a pool with these tokens it won't be added automatically so maybe it's better to link the tag to the specific token instead of the pool?

mendesfabio:
thanks @groninge01 I didn't know that was possible. These pools have been created by Gnosis/KPK and are incentivized so at this moment are the only ones we would like to highlight. I'm wondering if we changed metadata to tokens will this same PR still work? Not clear to me whether tagging tokens works for pools list filtering or only for banners/icons in pool page.

groninge01:
as long as the tag gets added to the pool(s) the filtering will work and any other stuff that looks at tags
i will check it with franz or jarek

groninge01 (approved):

PR #669, opened 2025-02-25
fix: points display on pool list & detail pages

groninge01:
the api added `'POINTS'` as tag for any pool with any points
this simplifies the pool filter and the points display on the list page


agualis (changes requested):
Filtering by points I don't get any result: 

https://mono-frontend-v3-git-fix-points-display-for-lis-f6fb71-balancer.vercel.app/pools?poolTags=POINTS

agualis (approved):

PR #701, opened 2025-03-07
Add custom error messages support via metadata

danielmkm:
The goal here is to allow the maxis the ability to add custom error messages to both the pool detail and add liquidity pages for special cased pools via the metadata repo. We additionally add support for an `allowAddLiquidity` flag that overrides the blocking of add liquidity when present in the metadata.

Screenshot 2025-03-07 at 09 42 49
Screenshot 2025-03-07 at 09 42 43


packages/lib/modules/pool/pool.helpers.ts
```diff
@@ -302,7 +303,12 @@ const shouldBlockV3PoolAdds = false
  * Returns true if we should block the user from adding liquidity to the pool.
  * @see https://github.com/balancer/frontend-v3/issues/613#issuecomment-2149443249
  */
-export function shouldBlockAddLiquidity(pool: Pool) {
+export function shouldBlockAddLiquidity(pool: Pool, poolMetadata?: PoolMetadata) {
```
agorer, line 306: Working on https://github.com/balancer/frontend-monorepo/issues/702 and that is related to this one and have a question: should we show tooltip reasons (getPoolAddBlockedReason) even when it is forced to be allowed by the metadata?

danielmkm: In instances where it is explicitly allowed, I would not show the reason, it should be covered by the custom pool warnings

PR #704, opened 2025-03-07
feat: helper function to set multiple token balances within an imperso…

agualis:
…nated hook


agorer (approved):

agorer:
@agualis just one question related to this. Whenever we are using an anvil fork, what happens with API data? Will it just return outdated data?

agualis:
> @agualis just one question related to this. Whenever we are using an anvil fork, what happens with API data? Will it just return outdated data?

Yeah, but I think the data to check in the E2E tests will be normally overwritten onchain.

PR #727, opened 2025-03-12
[#702] Refactor shouldBlockAddLiquidity + reasons

### LINKED ISSUE #702: Refactor shouldBlockAddLiquidity + reasons
We have a dangerous duplication in two different functions that should be probably one: 

`shouldBlockAddLiquidity` and `getPoolAddBlockedReason`

- [ ] - Add tests for uncovered cases
- [ ] - Refactor in a way that we don't duplicate the block rules

agorer:
Closes #702


groninge01:
i wonder if we need to show all the reasons in the tooltip when there are also alerts shown above?

image

agorer:
> i wonder if we need to show all the reasons in the tooltip when there are also alerts shown above?

Wouldn't it be the case that we need to show either all the reasons or none (if we think the alerts are enough to convey why the button is disabled). Maybe @uiuxxx can shed some light on the matter.

packages/lib/shared/utils/text-format.ts
```diff
@@ -0,0 +1,3 @@
+export function formatTextListAsItems(items: string[]): string {
+  return items.reduce((acc: string, item: string) => acc + (acc ? '\n' : '') + '- ' + item, '')
```
agualis, line 2: According to Groninge's screenshot, the alerts VS tootip items are displayed in reversed order.
```suggestion
  return items.reverse().reduce((acc: string, item: string) => acc + (acc ? '\n' : '') + '- ' + item, '')
```

agualis (approved):

PR #728, opened 2025-03-13
fix: add poolIds to swap simulation query key

agualis:


packages/lib/modules/swap/SwapProvider.tsx
```diff
@@ -414,7 +414,7 @@ export function _useSwap({ poolActionableTokens, pool, pathParams }: SwapProvide
       setPriceImpact(calcMarketPriceImpact(tokenInUsd, tokenOutUsd))
     } else if (simulationQuery.data) {
       setPriceImpact(undefined)
-      setPriceImpactLevel('unknown')
+      setPriceImpactLevel('low')
```
agorer, line 417: what does this change mean?

agorer (approved):

packages/lib/modules/swap/SwapProvider.tsx
agualis: This is setting the "default" price impact level when tokenInUsd or tokenOutUsd are not > 0 but the default value should be "low" like you can see in the [useState creation](https://github.com/balancer/frontend-monorepo/blob/fix/swapSimulationKeys/packages/lib/modules/price-impact/PriceImpactProvider.tsx#L10), if not, this could cause Price unknown alert flickering in some scenarios.

PR #750, opened 2025-03-19
fix: token image optimization

groninge01:
- allow overflow for logo-in-logo images
- round square images a bit


groninge01:
@uiuxxx already tested it and found nothing weird

agorer (approved):

PR #756, opened 2025-03-20
fix: do not show empty tooltip

groninge01:
- disable the empty tooltip below the button when there are no blocking reasons to show

image


agorer:
This is also used in `PoolMyLiquidity.tx`, so we should also probably change it there.

As a side note, a more general solution can be accomplished using the pattern explained in https://chakra-ui.com/docs/components/tooltip#setup where disabling the tooltip returns just the contents, but up to you.

groninge01:
> This is also used in `PoolMyLiquidity.tx`, so we should also probably change it there.
> 
> As a side note, a more general solution can be accomplished using the pattern explained in https://chakra-ui.com/docs/components/tooltip#setup where disabling the tooltip returns just the contents, but up to you.

hmm I thought I checked if there was a prop for it but I somehow didn't see it :(

agorer (approved):

PR #760, opened 2025-03-21
Issue 742

agualis:


agualis:
Already merged into https://github.com/balancer/frontend-monorepo/pull/520

PR #775, opened 2025-03-25
detect insufficient allowance

agorer:


packages/lib/modules/transactions/transaction-steps/transaction.helper.ts
```diff
@@ -0,0 +1,10 @@
+import { TransactionResult } from '../../web3/contracts/contract.types'
+import { ManagedResult } from './lib'
+
+export function resetTransaction(v: ManagedResult) {
```
agualis, line 4: Better naming this file `transaction.helpers.ts` in plural as that's the pattern we use.

PR #781, opened 2025-03-26
Added protocol filter

### LINKED ISSUE #779: Display Protocol filter when selected
Usually when a filter is selected, the item is placed as a chip under the "Liquidity pools" header of the pools list (which the user can close directly from that chip UI). 

In this example, there are two filters selected: "Weighted" and "v2", however only the "Weighted" chip is displayed. 

Image

The task:
- Add a protocol chip to the filtered items list, if anything but "All" is selected.
- e.g. If "v2" is selected, show a "v2" chip with a close icon (e.g. how we are displaying the "Weighted" chip in the example above.
- If the default "All" is selected, do not display a chip.

0xAnon0602:
Fixes #779 

image
image

### DIFF
--- a/packages/lib/modules/pool/PoolList/PoolListFilters.tsx
+++ b/packages/lib/modules/pool/PoolList/PoolListFilters.tsx
@@ -66,7 +66,7 @@ const SLIDER_STEP_SIZE = 100000
 
 export function useFilterTagsVisible() {
   const {
-    queryState: { networks, poolTypes, minTvl, poolTags, poolHookTags },
+    queryState: { networks, poolTypes, minTvl, poolTags, poolHookTags, protocolVersion },
   } = usePoolList()
   const [isVisible, setIsVisible] = useState(false)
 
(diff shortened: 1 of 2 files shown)


packages/lib/modules/pool/PoolList/PoolListFilters.tsx
```diff
@@ -350,13 +355,37 @@ export function FilterTags({
...
+          >
+            <Tag size="lg">
+              <TagLabel>
+                <Text fontSize="sm" fontWeight="bold">
+                  {protocolVersion === 1 ? 'CoW' : `v${protocolVersion}`}
+                </Text>
+              </TagLabel>
+              <TagCloseButton onClick={() => setProtocolVersion(null)} />
+            </Tag>
+          </motion.div>
```
groninge01, line 386: maybe now is the time to make this more DRY

groninge01 (approved):

PR #794, opened 2025-03-27
[#783] veBAL: Expired veBAL error

### LINKED ISSUE #783: veBAL: Expired veBAL error
Bug report:
- When the user's lock for their 'B-80BAL-20WETH' has expired, they should not have any veBAL. 
- Currently, the display shows small amounts like `<0.0001`

Image

Image

Task:
- Fix the display so that the user has zero veBAL when their lock has expired. 
- This is most likely an API error, so it may be best to coordinate with Franz to see why the API is returning the error. 
- This error is also displayed on the Vote page within the 'My veBAL' chip (pictured above).
- The display should look per the designs below:

Design:
- https://www.figma.com/design/Jrm12lbqVzLh1F46nWjdoK/zen-veBAL?node-id=1120-5988&t=DnRswGpo8MmX6Ket-1

Expired wallet to impersonate to see and test this issue:
- 0x6C8F61d3043f639F3404910A632cE0eA154b51d2

agorer:
Closes #783


packages/lib/modules/vebal/useVebalUserData.ts
```diff
@@ -24,7 +24,7 @@ export function useVebalUserData() {
   return {
     data,
     refetch,
-    isConnected,
+    isConnected, // FIXME: [JUANJO] we shouldn't return account data from here
```
agualis, line 27: Can you fix it already in this PR?

agualis (changes requested):

packages/lib/modules/vebal/useVebalUserData.ts
agorer: I thought it was used in multiple places, but seems like isConnected is only used once. Will change it here.

agualis (approved):

PR #796, opened 2025-03-28
chore: update packages dependabot

groninge01:
- bump @apollo/experimental-nextjs-app-support from 0.11.3 to 0.11.11
  - `ssrMode` is no longer needed
- bump turbo from 2.3.3 to 2.4.4
- bump vitest from 3.0.5 to 3.0.9
- bump @graphql-codegen/typescript-document-nodes from 4.0.9 to 4.0.15

### DIFF
--- a/packages/lib/shared/services/api/apollo.client.ts
+++ b/packages/lib/shared/services/api/apollo.client.ts
@@ -18,7 +18,6 @@ export function createApolloClient() {
   })
 
   return new ApolloClient({
-    ssrMode: typeof window === 'undefined',
     link:
       typeof window === 'undefined'
         ? ApolloLink.from([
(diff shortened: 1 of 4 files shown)


groninge01:
changes in the test-api are breaking those build rn

packages/lib/shared/services/api/apollo.client.ts
```diff
@@ -18,7 +18,6 @@ export function createApolloClient() {
   })
 
   return new ApolloClient({
-    ssrMode: typeof window === 'undefined',
```
agorer, line 21: why is this no longer needed?

agorer (approved):

packages/lib/shared/services/api/apollo.client.ts
groninge01: https://github.com/apollographql/apollo-client-nextjs/pull/358

it's a bit harder to find out *why* the option was removed though

groninge01: i guess it was never needed? https://github.com/apollographql/apollo-client-nextjs/issues/350

PR #821, opened 2025-04-01
[#788] show the max amount of vebal when locked for one year

### LINKED ISSUE #788: veBAL: Show veBAL balance if locked for 1 year
Currently, this tooltip contains mocked data. It should be replaced with the real number. 

Image

The number should be the amount of veBAL the user would have if they extended their lock to 1-year. e.g. take the amount of B-80BAL-20WETH that the user has and lock it for 1-year.

agorer:
Closes #788

### DIFF
--- a/packages/lib/modules/vebal/lock/form/VebalLockForm.tsx
+++ b/packages/lib/modules/vebal/lock/form/VebalLockForm.tsx
@@ -201,7 +201,7 @@ export function VebalLockForm() {
               </Card>
             </GridItem>
             <GridItem>
-              {/* TO-DO Potential min. weekly yield */}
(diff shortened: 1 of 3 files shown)


packages/lib/modules/vebal/lock/form/VebalLockForm.tsx
```diff
@@ -201,7 +201,7 @@ export function VebalLockForm() {
               </Card>
             </GridItem>
             <GridItem>
-              {/* TO-DO Potential min. weekly yield */}
+              {/* TO-DO (vebal) Potential min. weekly yield */}
```
agorer, line 204: Marked so could be filtered with the rest of vebal TODOs

groninge01 (approved):

PR #825, opened 2025-04-02
[#785] get veBAL balance onchain instead of from API

### LINKED ISSUE #785: veBAL: New veBAL is taking several minutes to register
Bug report:

After locking tokens on a fresh account, and getting veBAL for the first time, it takes several minutes for the system to update for the user's new veBAL. For example, this warning is incorrectly displayed to the user, stating that they have no veBAL (even after a few minutes). This could be quite concerning for users until the update is registered by the system. 

Image

Note: The system has registered that the user has some veBAL, because the first warning mentions that the user's veBAL will expire soon. However, that is not reflected in the 'My veBAL' balance in the chip and with the displayed 2nd warning. So, there seems to be a timing mismatch from when the system first recognizes veBAL to when it is finally displayed.

To reproduce:
- Lock the ve8020 LP token to get veBAL on a new wallet that hasn't had any veBAL before
- See how long it takes for the veBAL on the user's UI to update.

agorer:
Closes #785

### DIFF
--- a/packages/lib/modules/vebal/VebalStats/UserVebalStatsValues.tsx
+++ b/packages/lib/modules/vebal/VebalStats/UserVebalStatsValues.tsx
@@ -10,9 +10,10 @@ import { useVebalLockData } from '@repo/lib/modules/vebal/lock/VebalLockDataProv
 import { PRETTY_DATE_FORMAT } from '@repo/lib/modules/vebal/lock/duration/lock-duration.constants'
 import { AlertIcon } from '@repo/lib/shared/components/icons/AlertIcon'
 import { useUserAccount } from '../../web3/UserAccountProvider'
+import { formatUnits } from 'viem'
 
 export type VebalUserStatsValues = {
-  balance: string
+  balance: bigint
   rank: number
   percentOfAllSupply: BigNumber | undefined
   lockedUntil: string | undefined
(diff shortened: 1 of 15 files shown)


packages/lib/modules/vebal/vote/Votes/MyVotes/MyVotes.tsx
```diff
@@ -38,7 +39,10 @@ export function MyVotes() {
   const { isOpen, onOpen, onClose } = useDisclosure()
   const { isConnected } = useUserAccount()
 
-  const { loading: vebalUserDataLoading, myVebalBalance, noVeBalBalance } = useVebalUserData()
+  const { isLoading: vebalUserDataLoading, veBALBalance, noVeBALBalance } = useVebalUserData()
+
+  // FIXME: [JUANJO] everything needs a number here, but we should work with bigint
```
agorer, line 44: Wherever numbers are used for calculation we need to change to use either bigint or bignumber, but we should probably have some tests around to refactor this.

packages/lib/modules/vebal/cross-chain/CrossChainBoost.tsx
```diff
@@ -42,7 +43,7 @@ export function CrossChainBoost() {
 
   const [syncIsOpen, setSyncIsOpen] = useState(false)
 
-  const { myVebalBalance } = useVebalUserData()
+  const { veBALBalance } = useVebalUserData()
```
agualis, line 46: We are calling `formatUnits(veBALBalance, 18)` in many places. Why not exposing a new `veBalBalanceUnits` from `useVebalUserData`?

agualis (changes requested):

packages/lib/modules/vebal/cross-chain/CrossChainBoost.tsx
agorer: Will probably first wait till we change all the calculations to use the raw bigint value, and then check if it is used just for formatting. Probably a formatting function without side effects?

packages/lib/modules/vebal/VebalStats/UserVebalStatsValues.tsx
```diff
@@ -22,17 +23,14 @@ export type VebalUserStatsValues = {
 export function UserVebalStatsValues() {
   const { isConnected } = useUserAccount()
   const lockData = useVebalLockData()
```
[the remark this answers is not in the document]
agorer: Was already being used, but made it more explicit.

agualis (approved):

Issue #784, opened 2025-03-27
veBAL: Bug when locking the minimum time period


uiuxxx:
**Describe the bug** 

When attempting to lock to get veBAL, I get the following error: `Can only lock until time in the future`

![Image](https://github.com/user-attachments/assets/9b5f3153-afea-4948-b01c-ab393b3138b7)

See bug here: https://www.loom.com/share/7e63f11b06e24788bb3a3c2079cada96

**To Reproduce** 

- Impersonate address: 0xC3409E1961610533305b0F309D99939d5D49B082
- Lock tokens at: https://balancer.fi/vebal/manage/lock
- Enter an amount and don't change the minimum period.

**Expected behavior** 

- The minimum period, should set the minimum period where no errors will occur. I understand that this should be at least 7 days. It looks like it is trying to lock until 3 April, which should be 7 days. 
- I'm currently in Australia on a Thursday (voting deadline day), so it's possible that there is some time confusion close to the deadline that is causing this.
- Note: When I update the lock duration time period do 2 weeks, it works perfectly fine.

PR #830, opened 2025-04-04
fix: lock date calculations

agualis:
Fixes https://github.com/balancer/frontend-monorepo/issues/784

Also adds impersonator utility to reset fork.


packages/lib/modules/web3/impersonation/ImpersonateAccount.tsx
```diff
@@ -25,6 +25,9 @@ export function ImpersonateAccount() {
       >
         Connect
       </Button>
+      <Button aria-label="Undo" disabled={!isAddress(impersonatedAddress)} onClick={() => reset()}>
```
agorer, line 28: why undo and not reset?

agorer (approved):

packages/lib/modules/web3/impersonation/ImpersonateAccount.tsx
agualis: Updated.

PR #833, opened 2025-04-04
Fix pool chart date display

### LINKED ISSUE #419: Handle chart timezone
Chart time data comes in Zulu time but we're not handling that correctly with the user's timezone so date tags can appear off.

Image

MattPereira:
Fixes #419

### Summary
- timestamps from API are UTC midnight 
  - example: `1743750000` which is `Fri, 04 Apr 2025 00:00:00 GMT`
- echarts defaults to interpreting timestamp values in local timezone of user
  -  converts `1743750000` to `Thu Apr 03 2025 17:00:00 GMT-0700 (Pacific Daylight Time)`
- adjusted timestamp by the offset from UTC to account for echarts limitation

### DIFF
--- a/packages/lib/modules/pool/PoolDetail/PoolStats/PoolCharts/PoolChartsProvider.tsx
+++ b/packages/lib/modules/pool/PoolDetail/PoolStats/PoolCharts/PoolChartsProvider.tsx
@@ -360,7 +360,13 @@ export function _usePoolCharts() {
       }
     }
 
-    return processedElements
+    // timestamps from chartData are UTC midnight, but echarts defaults to interpreting timestamp values in local timezone, so adjust by offset to display correct day on chart
+    return processedElements.map(([timestamp, value]) => {
+      const date = new Date(timestamp * 1000)
+      const timezoneOffset = date.getTimezoneOffset() * 60 // in seconds
+      const adjustedTimestamp = timestamp + timezoneOffset
+      return [adjustedTimestamp, value]
+    })
   }, [
     activePeriod.value,
     activeTab.value,
--- a/packages/lib/modules/pool/PoolDetail/PoolStats/PoolCharts/PoolChartsProvider.tsx
+++ b/packages/lib/modules/pool/PoolDetail/PoolStats/PoolCharts/PoolChartsProvider.tsx
@@ -15,6 +15,7 @@ import { useTheme as useNextTheme } from 'next-themes'
 import { isCowAmmPool } from '../../../pool.helpers'
 import { PoolChartTab, usePoolChartTabs } from './PoolChartTabsProvider'
 import { useMandatoryContext } from '@repo/lib/shared/utils/contexts'
+import { alignUtcWithLocalDay } from '@repo/lib/shared/utils/time'
 
 const MIN_DISPLAY_PERIOD_DAYS = 30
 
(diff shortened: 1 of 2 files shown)


groninge01:
lgtm!

only comment i have is to maybe extract this as a helper to `time.ts`, you also get some consts then to use instead of the magic numbers

groninge01 (approved):

MattPereira:
@groninge01 thanks for the pointers! :pray: 

before merging just want to confirm this is the desired UX where totals for each day across all timezones flip over as of UTC midnight. otherwise users would have different totals for each day which would not make sense right?

I'm in calfornia where as of 5pm local time it starts showing a day ahead in the chart, which is similar to my experience on [tradingview](https://www.tradingview.com/)

image

groninge01:
> ...where as of 5pm local time it starts showing a day ahead on the chart

i think that was the original issue, i am utc+2 and this is what i see now which is incorrect
image

so that chart date needs to move back 24 hrs + my tz offset and it will show the last snapshot with yesterday's date
but if we do that for california time (just subtract the tz offset then) it would show (at utc midnight) the last snapshot with today's date (which might be less weird than showing it with the day ahead's date?)

```
export function alignUtcWithLocalDay(timestamp: number) {
  const timezoneOffset = mins(new Date().getTimezoneOffset()).toSecs()
  return (
    timestamp -
    (timezoneOffset < 0 ? Math.abs(timezoneOffset) + twentyFourHoursInSecs : timezoneOffset)
  )
}
```

groninge01:
@MattPereira can you check how my suggestion above looks for you throughout the day? if ok we'll just do it like that...

MattPereira:
@groninge01 your proposal knocks me back to a day behind which was my interpretation of complaint from #419 

image

MattPereira:
Test at 5pm california time with @groninge01 proposal shows zero volume for the current day

image

MattPereira:
> which might be less weird than showing it with the day ahead's date?

IMO showing the day ahead date from California as UTC hits midnight is industry standard, but no idea how things look from UTC+ zones :sweat_smile: 

image

image

groninge01:
> shows zero volume for my current day which is April 9

did you check the api response to see if it contained data? this is what i get now for california time

image

groninge01:
> I think showing the day ahead date from California as UTC hits midnight is industry standard

tradingview is showing live data, the snapshot are static
seen from utc the contain data from the day before
this is the latest snapshot for me:

```
{
    "id": "0x5c6ee304399dbdb9c8ef030ab642b10820db8f56000200000000000000000014-1744243200",
    "timestamp": 1744243200,
    "totalLiquidity": "22333397.82772193",
    "volume24h": "103652.5804805756",
    "fees24h": "217.0150678511709",
    "surplus24h": "0",
    "sharePrice": "3.437048991098134",
    "__typename": "GqlPoolSnapshot"
}
```

timestamp is Apr 10 12:00 AM UTC but the data is about Apr 9, so even for California i don't think the date should be the day ahead because that is not the day the data is about

MattPereira:
@groninge01 saw the [convo on slack with Jarek](https://balancerecosystem.slack.com/archives/C05H1NS8Q8L/p1744273891166849?thread_ts=1744180103.762519&cid=C05H1NS8Q8L). Is the plan to wait and have the API switch from 24 hour buckets to 1 hour buckets so each user sees different data for each day according to their timezone?

groninge01:
@MattPereira i just found out the latest snapshot is today's which will be updated untill a new snapshot is created at utc midnight
so this pr can be merged as it is now

PR #852, opened 2025-04-08
feat: show pool tokens for Beets

groninge01:
show tokens instead of names for pools (only commented out for now)

image


groninge01:
commented out for now because it might be needed in the future again or should i remove it?

PR #855, opened 2025-04-08
feat: add pool version filter to pool voting list

### LINKED ISSUE #802: veBAL: Add filters to VoteListFilters
Add protocol version to VoteListFilters (see PoolListFilters)
Then in FilterTags the props could be made mandatory.

And also hooks & minTVL?

agualis:
Closes #802 by adding pool protocol version filter to voting list. 

There are other potential filters (like hooks & `minTVL`) to be copied from pool list but I think this is enough. 

https://github.com/user-attachments/assets/724cd974-d179-498c-a73a-db7187239f6f

### DIFF
--- a/packages/lib/modules/vebal/vote/VoteList/VoteListFilters.tsx
+++ b/packages/lib/modules/vebal/vote/VoteList/VoteListFilters.tsx
@@ -22,6 +22,7 @@ import {
   FilterButton,
   PoolNetworkFilters,
   PoolTypeFilters,
+  ProtocolVersionFilter,
 } from '@repo/lib/modules/pool/PoolList/PoolListFilters'
 import { useVoteList } from '@repo/lib/modules/vebal/vote/VoteList/VoteListProvider'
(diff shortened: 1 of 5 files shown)


packages/lib/modules/vebal/vote/VoteList/VoteListFilters.tsx
```diff
@@ -124,6 +131,19 @@ export function VoteListFilters() {
                           toggledNetworks={toggledNetworks}
                         />
                       </Box>
+                      <Box as={motion.div} variants={staggeredFadeInUp}>
```
agorer, line 134: don't we have all this as a component on pool list filters?

agorer (approved):

packages/lib/modules/vebal/vote/VoteList/VoteListFilters.tsx
agualis: No we don't, we could extract them if we refactor the shared filter state but I decided not do it yet.

PR #872, opened 2025-04-10
feat: migrate to breaking changing sdk v3.1.0

agualis:
The last sdk version has important breaking changes as it exports addresses in a different way: 

Context: https://github.com/balancer/b-sdk/pull/644

### DIFF
--- a/packages/lib/config/networks/fantom.ts
+++ b/packages/lib/config/networks/fantom.ts
@@ -3,8 +3,6 @@ import { NetworkConfig } from '../config.types'
 import { zeroAddress } from 'viem'
 import { convertHexToLowerCase } from '@repo/lib/shared/utils/objects'
 import { emptyAddress } from '@repo/lib/modules/web3/contracts/wagmi-helpers'
-import { BALANCER_COMPOSITE_LIQUIDITY_ROUTER_BOOSTED, BALANCER_ROUTER } from '@balancer/sdk'
-import { fantom } from 'viem/chains'
 
 const networkConfig: NetworkConfig = {
   chainId: 250,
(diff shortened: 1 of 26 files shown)


packages/lib/config/networks/fantom.ts
```diff
@@ -43,8 +41,6 @@ const networkConfig: NetworkConfig = {
       vaultV2: '0x20dd72Ed959b6147912C2e529F0a0C651c33c9ce',
       relayerV6: '0x0faa25293a36241c214f3760c6ff443e1b731981',
       minter: zeroAddress,
```
agualis, line 43: Removing these ones as they are not yet defined.

packages/lib/modules/pool/queries/usePoolEnrichWithOnChainData.tsx
```diff
@@ -14,9 +14,11 @@ import {
 } from '../../web3/contracts/abi/generated'
 import { isComposableStablePool } from '../pool.utils'
 import { cowAmmPoolAbi } from '../../web3/contracts/abi/cowAmmAbi'
-import { weightedPoolAbi_V3, vaultExtensionAbi_V3 } from '@balancer/sdk'
+import { vaultExtensionAbi_V3 } from '@balancer/sdk'
 import { getCompositionTokens } from '../pool-tokens.utils'
 
+const totalSupplyAbi = parseAbi(['function totalSupply() view returns (uint256)'])
```
agualis, line 20: `weightedPoolAbi_V3` is not exposed by the SDK anymore so we now manually use the ABI function

groninge01 (approved):

packages/lib/modules/pool/queries/usePoolEnrichWithOnChainData.tsx
MattPereira: my bad we could have left this for you i didn't realize FE using it

MattPereira: if there are v3 ABIs FE uses that could change with new deployments would be easy to add back into SDK automation as needed/desired

agualis: No worries, it is ok to use `parseAbi`.

PR #904, opened 2025-04-15
Add EZKL banner

uiuxxx:
Add EZKL banner


packages/lib/modules/fee-managers/getFeeManagersMetadata.ts
```diff
@@ -0,0 +1,18 @@
...
+} from '@repo/lib/shared/utils/fetchAndLowercaseAddresses'
+
+const FEE_MANAGERS_METADATA_URL =
+  'https://raw.githubusercontent.com/balancer/metadata/refs/heads/main/fee_managers/index.json'
+
+export type FeeManagersMetadata = {
+  id: string
+  name: string
+  description: string
+  addresses: Record<string, string[]> // chainId -> addresses[]
```
agorer, line 13: Wouldn't a better type for this be something like:
```
`0x${string}`
```
we can even make an alias to make it clearer.

packages/lib/modules/fee-managers/useFeeManager.ts
```diff
@@ -0,0 +1,26 @@
+import { useFeeManagers } from './FeeManagersProvider'
+import { getChainId } from '@repo/lib/config/app.config'
+import { Pool } from '../pool/pool.types'
+
+export function useFeeManager(pool: Pool) {
```
agorer, line 5: Don't know anything about the fee manager topic. Can it only be associated with a chain or does specific pools can have their own fee manager?

packages/lib/modules/fee-managers/useFeeManager.ts
```diff
@@ -0,0 +1,26 @@
...
+  const { metadata } = useFeeManagers()
+
+  const hasFeeManager = !!pool.swapFeeManager
+  const chainId = getChainId(pool.chain)
+
+  const feeManager = metadata?.find(metadata => {
```
agorer, line 11: Having both elements be named metadata is a little confusing.

packages/lib/modules/pool/PoolDetail/PoolBanners/PoolFeeManagerBanner.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { usePool } from '../../PoolProvider'
+import { EzklPromoBanner } from '@repo/lib/shared/components/promos/EzklPromoBanner'
+
+export function PoolFeeManagerBanner() {
+  const { pool } = usePool()
+  const { hasFeeManager, feeManager } = useFeeManager(pool)
+
+  if (!hasFeeManager) return null
+
+  if (feeManager?.id === 'fee_manager_ezkl') {
```
agorer, line 11: Maybe using a constant here that could be shared in other parts of the code could be better?

agorer (approved):

packages/lib/modules/pool/PoolDetail/PoolBanners/PoolFeeManagerBanner.tsx
groninge01: The thought crossed my mind but i think this will be the only place it will be used.
Can always revisit if it's needed anywhere else.

packages/lib/modules/fee-managers/useFeeManager.ts
groninge01: pools can optionally have their own swap fee manager, it is not chain specific

PR #943, opened 2025-04-24
Fix animation on list filters popovers

agorer:
https://github.com/user-attachments/assets/678ed143-afb4-40a2-8c8b-f6394adb8deb


groninge01:
PortfolioFilters?

agorer:
> PortfolioFilters?

You are right, didn't know those existed. After examination though, right now they are not losing frames (probably due to being less complicated).

PR #958, opened 2025-04-25
[#939] Change veBAL locks chart tooltip styles

### LINKED ISSUE #939: Bug: veBAL amount in "Manage veBAL" chart is repeated
A some positions, the veBAL amount in the tooltip is incorrectly repeated:

Image

https://www.loom.com/share/53a48cf7b707460d98634a906e6632bb

agorer:
Closes #939 

@uiuxxx I don't think it is possible to have a tooltip that it is continuous (eChart seems to only show tooltips for data points) but I have added interpolated points where small density of data points (right now if there is more than 30 days between locks / data points a new interpolated data point will be added every 7 days to fill the void).


groninge01:
you could get rid of `/|* 1000` by using the date-fns conversion helpers

groninge01 (approved):

uiuxxx:
@agorer This is great! The 7 day data points works fine and is much better than before.

PR #959, opened 2025-04-28
chore: update sdk to 3.2.0

agualis:


apps/frontend-v3/package.json
```diff
@@ -25,8 +25,7 @@
   },
   "dependencies": {
     "@apollo/client": "^3.13.1",
-    "@balancer/sdk": "3.1.0",
-    "bignumber.js": "9.2.0",
+    "@balancer/sdk": "3.2.0",
```
brunoguerios, line 28: Can you please source from my working branch on b-sdk?
```suggestion
    "@balancer/sdk": "github:balancer/b-sdk#bump-deps-latest",
```
And then please double check fi the app is working as expected.
I don't expect any specific issue, so it's more like a general overview that nothing is broken after the update.

PR #974, opened 2025-04-30
fix: remove useGaugeVotes duplication

agualis:
`VotesProvider` is wrapping `VoteListProvider` so it makes sense to expose gaugeVotes to avoid duplicating `useGaugeVotes` call.


agorer (approved):

agorer:
Why are there duplicated commits in this PR?

Issue #986, opened 2025-05-05
Min-TVL filter improvements


uiuxxx:
Currently users can either have $0 Min-TVL with the next lowest being $100k due to how the slider works. 

Let's improve this by adding a number input that complements the slider. The main tasks:
- Add a number input alongside the slider
  - <input type="number">  (so user's can't add any non-numeric characters)
  - Set min = 0. No max needed.
  -  Ideally reformat the number after the input loses focus to show commas within the numbers for better legibility. 
- The slider may need a few changes to match
  - It will still need a range: Let's make it 0 - $100m
  - If the user sets a number higher than $100m, we have some options that you can investigate:
    - Disable the slider while in the maxed position, forcing the user to interact with the input to bring the number back down. This will avoid an unexpected slider experience which 
    - Reset the scale to the number in the input box.
    - If both options are complicated, we could consider totally removing the slider completely, since the number input covers all cases by itself.

Designs:

https://www.figma.com/design/xiTeGFg2XxRbcM2t00b7E2/v3-app-2025?node-id=1513-541&t=j6QA0K1AS4ELcT7b-1

![Image](https://github.com/user-attachments/assets/5897f41a-a731-40d9-ad68-2b4e4c6849a5)

groninge01:
fixed by #987

PR #987, opened 2025-05-05
feat: remap tvl values to slider

groninge01:
alternative solution for #986


uiuxxx:
Reviewed it on `mono-test-v3` while the frontend was building. Looks and works great!

agualis (approved):

PR #1001, opened 2025-05-06
feat: upgrade eslint to v9

groninge01:
- upgrade eslint to v9 (w/ conversion to new configs)


apps/beets-frontend-v3/lib/modules/reliquary/ReliquaryProvider.tsx
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable max-len */
+/* eslint-disable react-hooks/rules-of-hooks */
```
agualis, line 1: This looks dangerous. Cannot we fix the root cause?

groninge01: fixed!

agualis (approved):

agualis (approved):

PR #1011, opened 2025-05-08
spike: migrate from rainbowkit to reown AppKit

agualis:
### Blockers

Dev only works in webpack mode (`pnpm run dev:webpack`) due to [this issue](https://github.com/reown-com/appkit/issues/4310)

### TODOs
- Review reown theming (there are only a bunch of options)
- Decide which wallets to include in the featured list

### Nice to haves
- Use custom disconnection modal instead of default reown modal?
Screenshot 2025-05-08 at 18 49 20 

- Use custom switch network modal instead of the default reown modal?

Screenshot 2025-05-08 at 19 05 32

If not, can we change the avatar to be the same as the one in the `ConnectWallet` button?

- The create coinbase wallet button was removed. I don't think we really need it but it would not be difficult to implement it if need it

### Before deployment:
- Test connection features (including safe app and safe connection though wallet connect)
- [optional] Elaborate a communication plan to avoid users scared about change:
  - pre-deployment banner announcing that wallet modal is going to be changed soon?
  - post-deployment banner to confirm the change?
  - social networks announcement?


agorer:
They seem to have started updating Rainbow kit again so we can probably continue using it for now (https://github.com/balancer/frontend-monorepo/pull/1414)

PR #1030, opened 2025-05-14
fix voting deadline calculation

agorer:
didn't work on timezones to the left of UTC


agualis (approved):
Can you extract it to a function with a unit tests covering before/after UTC timezones?

(note that all tests run with Madrid TZ by default)

PR #1038, opened 2025-05-16
[#790] Vote optimization

### LINKED ISSUE #790: veBAL: Optimize incentives algorithm
**User story:**

As a user, I want to see how much I can make from voting in an optimal way to get the maximum incentives.

**Solution:**

- Show users the total amount of their potential incentives for the week if the vote in an optimized fashion. 
- Allow the user's to apply this with a single click

Image

**Details:**
- When a user has a connected wallet with veBAL > 0, we need to determine their optimal voting strategy. 
- The system should display the total number of incentives the user may receive if they vote in the optimal way. 
- If the user taps the 'Apply' button, the system should auto-fill the pools into the 'My votes' table and also auto-complete the vote percentages. 

**Algorithm**
- The most basic version is to optimize votes based on the user's veBAL holdings, when they have no previous votes. 
- Ideally, our algorithm will also take into account timelocked vote. For example if a user has 20% of their votes allocated to a pool that is timelocked with 80% of their votes eligible to be allocated. Our algorithm should do the calculation to only allocate the 80% remaining, and exclude any pool where the user has timelocked votes. 

**References:**
- A lot of the logic for this already exists in https://defilytica.tools/#/balancer/veBALMultiVoter
- I understand that defilytica doesn't take timelocked votes into account.
- It's best to talk with Xeonus and Daniel for more details.

**Related tickets:**

- This ticket solely relates to the algorithm. 
- This ticket should be done in conjunction with #791 (which includes more details on the remain states, like when a user has an unconnected wallet).

agorer:
Closes #790


apps/frontend-v3/lib/vebal/vote/Votes/MyVotes/MyVotesStats/incentives-optimization/useIncentivesOptimized.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+      totalVotes,
+      blacklistedVotes,
+      inputsLoading
+    )
+  )
+
+  return result
+}
+
+describe('Incentives optimization', () => {
```
agualis, line 35: 🙌

apps/frontend-v3/lib/vebal/vote/Votes/MyVotes/MyVotesStats/incentives-optimization/useIncentivesOptimized.ts
```diff
@@ -0,0 +1,262 @@
...
+  gaugeAddress: Address
+  votes: BigNumber
+  incentivesAmount: number
+  userVotes: BigNumber
+  userPrct: BigNumber
+}
+
+const PERCENT_STEP = bn(0.0001)
+
+export function useIncentivesOptimized(
```
agualis, line 30: ❤️‍🔥

agualis (approved):

agualis:
Great work 🎩

PR #1067, opened 2025-05-22
fix: wagmi provider error in apps

groninge01:


packages/lib/shared/utils/wagmi.ts
```diff
@@ -0,0 +1,19 @@
+// this barrel file is used when the apps need to call hooks from wagmi
```
agualis, line 1: Add more context for future developers to understand why this is needed.

agualis: Also, in order to avoid regressions I would setup `eslint.config.js` in bal and beets apps with this: 

```
import nextJsConfig from '@repo/eslint-config/next-js'

/** @type {import('@typescript-eslint/utils').TSESLint.FlatConfig.ConfigFile} */
export default [
  {
    ignores: [
      // Ignore dotfiles
      '.*.js',
      '.next/**',
      '.turbo/**',
      'v8-compile-cache-0/**',
      'next-env.d.ts',
      'next.config.js',
      'node_modules/**',
    ],
  },
  ...nextJsConfig,
  {
    rules: {
      // App-specific overrides can be added here
      'no-restricted-imports': [
        'error',
        {
          patterns: [
            {
              // Disable direct wagmi imports from balancer and beets apps to avoid useWagmi config error (introduced by eslint + nextjs 15)
              group: ['wagmi'],
              importNames: ['useReadContract'],
              message: 'Import from @repo/lib/shared/utils/wagmi instead',
            },
          ],
        },
      ],
    },
  },
]
```

so that the linter makes sure that we do have direct wagmi imports from those apps.

agualis (changes requested):

agualis (approved):

Issue #1027, opened 2025-05-14
Ledger + MM signature error


agualis:
Some users are facing this issue when trying to run transactions using Ledger + Metamask:

https://discord.com/channels/638460494168064021/1356279093045498029/1356279506733895811

agorer:
<img width="725" alt="Image" src="https://github.com/user-attachments/assets/5cdde573-c114-41d5-bc5c-cee361a4d2b8" />

PR #1068, opened 2025-05-22
fix: use increaseApproval instead of approval when approving veBal Bpt

agualis:
Closes [#1027](https://github.com/balancer/frontend-monorepo/issues/1027)

This PR only uses the alternative `increaseApproval` instead of `approval` for `veBAL BPT` to workaround an issue that Ledger users are facing in combination with wallets like Metamask.

After removing the approval step in the increase veBAL lock [in this PR](https://github.com/balancer/frontend-monorepo/pull/1087) that won't be an issue anymore, so this PR can be closed. 

However, Ledger users can still face issues when [approving V2 BPTs for staking ](https://github.com/balancer/frontend-monorepo/blob/020b84d5545d66f1c49db60032f19157daad9237/packages/lib/modules/pool/actions/stake/useStakeSteps.tsx#L19). After [this PR](https://github.com/balancer/frontend-monorepo/pull/1074) they will see a custom error message with some recommendations.

Adding a proper fix in `useTokenApprovalSteps` is possible but tricky so maybe we want to wait until we understand the `increaseApproval` edge case better. 

cc @agorer @groninge01

### DIFF
--- a/packages/lib/modules/web3/contracts/abi/vebalAbi.ts
+++ b/packages/lib/modules/web3/contracts/abi/vebalAbi.ts
@@ -1,0 +1,20 @@
+import { AbiItem, erc20Abi } from 'viem'
+
+const increaseApprovalAbiItem: AbiItem = {
+  inputs: [
+    { internalType: 'address', name: 'spender', type: 'address' },
+    { internalType: 'uint256', name: 'amount', type: 'uint256' },
+  ],
+  name: 'increaseApproval',
+  outputs: [{ internalType: 'bool', name: '', type: 'bool' }],
+  stateMutability: 'nonpayable',
+  type: 'function',
+} as const
+
+/*
+  Edge case:
+  when locking veBalBPT for getting veBal, the veBalBPT contract must use its increaseApproval function
+  instead of the standard Erc20 approve function.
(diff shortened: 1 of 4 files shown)


packages/lib/modules/web3/contracts/abi/vebalAbi.ts
```diff
@@ -0,0 +1,20 @@
...
+  outputs: [{ internalType: 'bool', name: '', type: 'bool' }],
+  stateMutability: 'nonpayable',
+  type: 'function',
+} as const
+
+/*
+  Edge case:
+  when locking veBalBPT for getting veBal, the veBalBPT contract must use its increaseApproval function
+  instead of the standard Erc20 approve function.
+  Here we extend the erc20Abi with the increaseApproval function to cover that edge case in the generic token approval flow.
```
agorer, line 18: Maybe explain the wallet combination that didn't work.

agorer (approved):

packages/lib/modules/web3/contracts/abi/vebalAbi.ts
agualis: I will add it when we have more info about it.

PR #1106, opened 2025-05-27
refactor: persist lbp creation config in local storage

### LINKED ISSUE #1105: Save pool config in local storage
We want to persist the information user enters in case of page refresh

See `UserSettingsProvider.tsx` for example of how we want to save data to local storage

MattPereira:
Closes #1105 

### Summary
- `usePersistentForm` hook to combine react hook form with local storage persistence
- persisted lbp config including step index tracking
- followed existing local storage key pattern


packages/lib/modules/lbp/LbpFormProvider.tsx
```diff
@@ -3,9 +3,12 @@
 import { useSteps } from '@chakra-ui/react'
 import { useMandatoryContext } from '@repo/lib/shared/utils/contexts'
 import { PropsWithChildren, createContext } from 'react'
-import { useForm } from 'react-hook-form'
+import { usePersistentForm } from '@repo/lib/shared/hooks/usePeristedForm'
```
agorer, line 6: Typo on the file name

packages/lib/shared/hooks/usePeristedForm.ts
```diff
@@ -0,0 +1,38 @@
+import { useLocalStorage } from 'usehooks-ts'
```
agorer, line 1: I have been thinking lately about adding more tests to the app (trying to refactor things around seems like a minefield sometimes). This looks like a good candidate to add some tests, but up to you.

agorer (approved):

packages/lib/shared/hooks/usePeristedForm.ts
MattPereira: agree with the sentiment but want to explore building the transaction modal first to scope out the difficult parts
- #1128

Issue #653, opened 2025-02-21
Add sale structure inputs


garethfuller:
- [x] Sale period inputs
- [x] Mechanism inputs
- [x] Seed liquidity inputs

PR #1132, opened 2025-05-29
[#653] Sale structure form improvements

agorer:
- Allow for custom weight adjustment
- Preview of token icon
- Get help button to open the learn more modal
- Button to paste the token address on the input

Still not 1:1 with the designs (missing some error management) but all the basic pieces and behaviour are there.

### DIFF
--- a/packages/lib/modules/lbp/LbpFormProvider.tsx
+++ b/packages/lib/modules/lbp/LbpFormProvider.tsx
@@ -30,6 +30,8 @@ export function useLbpFormLogic() {
       endTime: '',
       collateralTokenAddress: '',
       weightAdjustmentType: 'linear_90_10',
+      customStartWeight: 90,
+      customEndWeight: 10,
       saleTokenAmount: '',
       collateralTokenAmount: '',
     },
--- a/packages/lib/modules/lbp/steps/WeightAdjustmentTypeInput.tsx
+++ b/packages/lib/modules/lbp/steps/WeightAdjustmentTypeInput.tsx
@@ -1,0 +1,161 @@
+import { Control, Controller, UseFormSetValue, UseFormWatch } from 'react-hook-form'
+import { SaleStructureForm } from '../lbp.types'
+import { Box, HStack, Stack, Text, VStack } from '@chakra-ui/react'
+import { ArrowRight } from 'react-feather'
+import { SelectInput } from '@repo/lib/shared/components/inputs/SelectInput'
+import { GqlChain } from '@repo/lib/shared/services/api/generated/graphql'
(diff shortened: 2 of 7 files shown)


agorer:
> Not sure if should be taken care of as part of this PR but noticed z-index issue w/ the dropdowns

I saw it, but it is from the dropdown component we are using. Tried changing the content z-index but it is not that simple. The lib page recommend using portals but it didn't work either. I will try to fix it some other way when I come back to fix the error management, but if nothing else works we can always hack it by hiding the other combo when opening the first one.

packages/lib/modules/lbp/steps/WeightAdjustmentTypeInput.tsx
```diff
@@ -0,0 +1,155 @@
...
+          <Box textAlign="right" w="100px">
+            <Text
+              fontSize="sm"
+              textColor="font.secondary"
+            >{`${collateralTokenSymbol}: ${100 - customWeight}%`}</Text>
+          </Box>
+        </HStack>
+        <SliderWithSteps
+          aria-label="weights-slider"
+          onChange={value => setValue('customStartWeight', value)}
```
MattPereira, line 147: looks like hard coding `customStartWeight` here is preventing to possibility to edit the ending weights

https://www.loom.com/share/440e35061dee4e639fbb29e3233f111e?sid=c539b21f-e255-4daf-83af-bd35d9b2a5ec

packages/lib/modules/lbp/LbpFormProvider.tsx
```diff
@@ -30,6 +30,8 @@ export function useLbpFormLogic() {
       endTime: '',
       collateralTokenAddress: '',
       weightAdjustmentType: 'linear_90_10',
+      customStartWeight: 90,
+      customEndWeight: 10,
```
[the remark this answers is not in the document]
agorer: This is only the projectTokenStartWeight / projectTokenEndWeight in case of custom weights. The calculation will be something like:

```
const projectTokenStartWeight = weightAdjustmentType === 'linear_90_10' ? 90 : ... : customStartWeight
```

packages/lib/modules/lbp/steps/WeightAdjustmentTypeInput.tsx
agorer: You are right, fixed.

MattPereira (approved):
good to be working with u ser :saluting_face:

agorer:
> side note that we do eventually need to figure out how to limit min weight to 1% in both directions or tx will revert

The component allows for min and max, I have changed it.

PR #1133, opened 2025-05-30
refactor: move onSuccess tx handling to global provider

agualis:
After the previous trasaction handling refactors, TransactionStateProvider was not needed anymore but I reused it to globally handle the "onSuccess" state by step txhash. This will prevent unnecessary refetch calls (that we were sometimes avoiding with useMemos and useCallbacks that will become now unnecessary).

### DIFF
--- a/packages/lib/modules/transactions/transaction-steps/lib.tsx
+++ b/packages/lib/modules/transactions/transaction-steps/lib.tsx
@@ -134,7 +134,7 @@ type TransactionInput =
   | ManagedSendTransactionInput
 
 // In some cases, the onSuccess callback must be retried, for instance, when the refetchApprovals has not finished yet
-export const RetryOnSuccess = 'retry'
+export const Retry = 'retry'
 
 export type TransactionStep = {
   id: string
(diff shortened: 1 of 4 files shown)


packages/lib/modules/transactions/transaction-steps/lib.tsx
```diff
@@ -133,6 +133,9 @@ type TransactionInput =
   | ManagedErc20TransactionInput
   | ManagedSendTransactionInput
 
+// In some cases, the onSuccess callback must be retried, for instance, when the refetchApprovals has not finished yet
+export const RetryOnSuccess = 'retry'
```
agorer, line 137: with that name it seems like it will retry if success happens, wouldn't retry be enough?

agorer (approved):

packages/lib/modules/transactions/transaction-steps/lib.tsx
agualis: 👍 Renamed to `Retry`

PR #1152, opened 2025-06-04
[LBP] feat: lbp header stats

groninge01:


packages/lib/modules/pool/LbpDetail/LbpHeader/LbpHeaderStats.tsx
```diff
@@ -0,0 +1,64 @@
...
+  label: string
+  value?: string | number | undefined | null
+}
+
+const IMAGE_TRANSFORM_ARRAY = ['rotate(180deg) scale(2)', 'rotate(180deg)', 'scale(1)']
+
+export function LbpHeaderStats() {
+  const stats: StatItem[] = [
+    {
+      label: 'Funds raised',
```
agorer, line 16: Does this mean we still don't have this info? Shouldn't we add a TODO if that is the case?

agorer (approved):

PR #1186, opened 2025-06-10
veBAL improvements

uiuxxx:
- Fix calendar bug which disabled today and future days before the vote deadline
- New section on the `vebal/manage` page to house cross-chain sync and claims
- Better hover states for the pools listed in the veBAL voting pages
- Prevent no veBAL alert bar from displaying until data is loaded.
- Reorder incentives icon in MyVotes stats

Screenshots:

New layout for cross-chain sync and claim

cs 2025-06-11 at 17 47 54@2x

Calendar highlighting today's date:

cs 2025-06-11 at 17 48 01@2x


apps/frontend-v3/lib/vebal/vote/VoteList/VoteListTable/VoteListTableRow.tsx
```diff
@@ -68,7 +68,7 @@ export function VoteListTableRow({ vote, keyValue, ...rest }: Props) {
             <NetworkIcon chain={vote.chain} size={6} />
           </GridItem>
           <GridItem>
-            <Link href={getPoolPath(vote)} target="_blank">
+            <Link href={getPoolPath(vote)} target="_blank" role="group">
```
agorer, line 71: Why are we using this role here? Keep in mind that changing the role has accessibility implications as well as testing ones (for testing functions that rely on the role are used to perform actions).

agorer (approved):

apps/frontend-v3/lib/vebal/vote/VoteList/VoteListTable/VoteListTableRow.tsx
uiuxxx: Previously with Chakra, I think you needed to use `role="group"` to enable `_groupHover` interactions from the parent element. 

cs 2025-06-12 at 11 36 47@2x

However, it now looks like you can also just use `className="group"` on the parent element. 

I have now updated this.

PR #1218, opened 2025-06-16
Fix LBP permit signature step

### LINKED ISSUE #1192: LBP creation: fix permit signature bug
When user first sees tx stepper, step 5 only registers one of the tokens for the signature creation, which breaks init 

Image

if you refresh the page, it is fixed

if you add console.log to see the init pool amounts that are fed as arg to the hook that creates permit2 signatures, it is fixed

MattPereira:
Closes #1192

### Summary
- Fuck `useMemo`
- ~~Somehow `WETH` does not make it into `details.batchApprovalTokens` for the [useSignPermit2InitializeStep](https://github.com/balancer/frontend-monorepo/blob/main/packages/lib/modules/pool/actions/initialize/useSignPermit2InitializeStep.tsx) but both tokens are there in the details of [useSignPermit2](https://github.com/balancer/frontend-monorepo/blob/main/packages/lib/modules/transactions/transaction-steps/useSignPermit2Step.tsx) hook~~
- ~~I have verified the bug also happens if you use `USDC` or `USDT` for the collateral token so I do not believe issue related to `wethIsEth`~~
- ~~page refresh or adding console.log AFTER running through the config steps seems to fix 100% of the time. However, saving console.log BEFORE running through config does not fix~~

Screenshot from 2025-06-16 13-51-50

### DIFF
--- a/packages/lib/modules/lbp/steps/useCreateLbpSteps.tsx
+++ b/packages/lib/modules/lbp/steps/useCreateLbpSteps.tsx
@@ -37,7 +37,7 @@ export function useCreateLbpSteps() {
   const signPermit2Step = useSignPermit2InitializeStep({ initPoolInput })
   const initLbpStep = useInitializeLbpStep({ initPoolInput })
 
-  const isSignPermit2Loading = (signPermit2Step?.details?.batchApprovalTokens?.length ?? 0) < 2
+  const isSignPermit2Loading = !signPermit2Step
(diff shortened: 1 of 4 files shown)


agorer (approved):
Is this the same step used on swaps, etc? If so, it would be nice to check that nothing has broken (the only thing I can think of is that sometimes infinite rendering loops are created).

MattPereira:
@agorer we could just keep the `useMemo` and add `details` to dependency array

PR #1224, opened 2025-06-17
UI tweaks: Better price impact labels and tooltips on swaps

uiuxxx:
- Better price impact label and tooltips for swaps
- Making the swap summary styles consistent with add liquidity
- Fix various minor UI bugs
  - 'Pool name' label on Pool list table header
  - Double divider on Pool transactions table
  - Better padding between stats in Pool page Stats section
  - Consistent sizing of transaction settings across swaps and add liquidity forms

### Screenshots:

New label and tooltip for swap 'price impact' 

cs 2025-06-17 at 14 54 07@2x

### DIFF
--- a/apps/frontend-v3/wagmi.config.ts
+++ b/apps/frontend-v3/wagmi.config.ts
@@ -82,6 +82,16 @@ export default defineConfig(() => {
           },
         ],
       }),
+      etherscan({
+        apiKey: env.ETHERSCAN_API_KEY,
+        chainId: 8453,
+        contracts: [
+          {
+            name: 'ReClammPool',
+            address: '0x03D5bbf2711926Ea3d7f2a5a07C8E5C6b25A1745',
+          },
+        ],
+      }),
     ],
   }
 })
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "stylelint": "16.14.1",
     "stylelint-config-standard": "^36.0.1",
     "stylelint-prettier": "^5.0.3",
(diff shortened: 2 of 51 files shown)


packages/lib/modules/eclp/hooks/EclpChartProvider.tsx
```diff
@@ -25,14 +33,27 @@ export function useEclpChartLogic() {
...
+  const { theme: nextTheme } = useNextTheme()
   const tokens = useMemo(() => {
     const poolTokens = getPoolActionableTokens(pool).map(token => token.symbol)
 
     return isReversed ? poolTokens.reverse().join('/') : poolTokens.join('/')
   }, [pool, isReversed])
 
-  const secondaryFontColor = theme.semanticTokens.colors.font.secondary.default
+  const secondaryFontColor =
+    nextTheme === 'dark'
```
agorer, line 44: we can probably create a  helper function to select colors that we can use around. Something like:

```
function useSelectColor() {
  const theme = useChakraTheme()
  const { theme: nextTheme } = useNextTheme()
  return (element:string, attr: string) => nextTheme === 'dark'
      ? theme.semanticTokens.colors[element][attr]._dark
      : theme.semanticTokens.colors[element][attr].light ? theme.semanticTokens.colors[element][attr].light
      : theme.semanticTokens.colors.[element].[attr].default
}
```
this will return a function that we can use in most other cases in the app.

agorer (approved):

PR #1226, opened 2025-06-18
[#1118] My Purchases and My Transactions on LBP pool detail

### LINKED ISSUE #1118: LBP detail page: my purchases/transactions

agorer:
Closes #1118


groninge01:
- could we break up `MyTransactions` into a few smaller re-useable components?
- at least using the same layout would be benificial i think

image

groninge01:
> > * could we break up `MyTransactions` into a few smaller re-useable components?
> > * at least using the same layout would be benificial i think
> 
> Not sure I am following, what do you mean?

never mind it for now

groninge01 (approved):

PR #1229, opened 2025-06-19
[LBP] make x handle optional

groninge01:
x handle is optional in the api so making it optional in the ui too

### DIFF
--- a/packages/lib/modules/lbp/steps/ProjectInfoStep.tsx
+++ b/packages/lib/modules/lbp/steps/ProjectInfoStep.tsx
@@ -206,7 +206,7 @@ function ProjectXHandle() {
 
   return (
     <VStack align="start" w="full">
-      <Text color="font.primary">X / Twitter handle</Text>
+      <Text color="font.primary">X / Twitter handle (optional)</Text>
       <Controller
         control={control}
         name="xHandle"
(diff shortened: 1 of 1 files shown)


agorer:
Have you talked with Pon about this one? In the designs there is even a connection button to check not only that it exists but also that it belongs to the user

groninge01:
> Have you talked with Pon about this one? In the designs there is even a connection button to check not only that it exists but also that it belongs to the user

i haven't but that still doesn't make it required
only when you fill in the handle we want to make sure it's the users handle right?

agorer:
I was just thinking that if it is that important to make the user login to their account maybe it is because it is something really important for the LBP (like the main channel of communication). But I don't really have much idea about this.

uiuxxx:
> > Have you talked with Pon about this one? In the designs there is even a connection button to check not only that it exists but also that it belongs to the user
> 
> i haven't but that still doesn't make it required only when you fill in the handle we want to make sure it's the users handle right?

I'm fine with X / Twitter not being a required field. 

But yeah, ideally, if it's not too much additional work, it would be great to verify ownership of the account, to give the public more confidence that the LBP is connected with a legitimate Twitter handle. We could of course do this for each social account, but I feel like X is probably the most important in the short term.

agorer (approved):

PR #1230, opened 2025-06-19
[#1202] Check that the image URL for the the launch token exists

### LINKED ISSUE #1202: Check that the logo added for the launch token exists
- If logo does not exists the data shouldn't be sent to the API

Image

- Also check that the project web exists

agorer:
Closes #1202

### DIFF
--- a/packages/lib/modules/lbp/steps/ProjectInfoStep.tsx
+++ b/packages/lib/modules/lbp/steps/ProjectInfoStep.tsx
@@ -1,6 +1,16 @@
-'use client'
-
-import { Heading, VStack, Text, HStack, Spacer, Divider, Checkbox, Button } from '@chakra-ui/react'
+import {
+  Heading,
+  VStack,
+  Text,
+  HStack,
+  Spacer,
+  Divider,
+  Checkbox,
+  Button,
+  InputGroup,
+  InputRightElement,
+  Spinner,
+} from '@chakra-ui/react'
 import { useLbpForm } from '../LbpFormProvider'
 import { ProjectInfoForm } from '../lbp.types'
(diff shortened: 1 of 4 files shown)


packages/lib/modules/lbp/steps/ProjectInfoStep.tsx
```diff
@@ -137,27 +151,42 @@ function TokenIconInput() {
...
-            onChange={e => field.onChange(e.target.value)}
-            placeholder="https://yourdomain.com/token-icon.svg"
-            value={field.value}
-          />
+          <InputGroup>
+            <InputWithError
+              error={isChecking ? '' : errors.tokenIconUrl?.message}
+              isInvalid={!isChecking && !!errors.tokenIconUrl}
+              onChange={e => field.onChange(e.target.value)}
+              placeholder="https://yourdomain.com/token-icon.svg"
```
groninge01, line 177: change to `.png`

groninge01:
maybe the easiest way is to let the user check if the image loads (add a info icon there or smth)?

agorer:
> maybe the easiest way is to let the user check if the image loads (add a info icon there or smth)?

Not sure I understand, wdym?

agorer:
> maybe at least do a check if the url is valid before checking anything else?
> 
> image

The URL in the image is valid, the problem is that the image does not exist (the validation for that field checks that the URL is valid at `url.hooks.ts`). I have added a check for showing the image only when no validation errors for the field.

groninge01:
> > maybe the easiest way is to let the user check if the image loads (add a info icon there or smth)?
> 
> Not sure I understand, wdym?

image

@uiuxxx will probably have a better suggestion

PR #1245, opened 2025-06-24
chore: bump packages

groninge01:
- bump @playwright/test to 1.53.1
- bump turbo to 2.5.4
- bump @eslint/js to 9.29.1

### DIFF
--- a/apps/beets-frontend-v3/package.json
+++ b/apps/beets-frontend-v3/package.json
@@ -52,7 +52,7 @@
     "@repo/typescript-config": "workspace:*",
     "@types/js-cookie": "^3.0.6",
     "@types/lodash": "^4.17.17",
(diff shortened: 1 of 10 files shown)


apps/beets-frontend-v3/package.json
```diff
@@ -52,7 +52,7 @@
     "@repo/typescript-config": "workspace:*",
     "@types/js-cookie": "^3.0.6",
     "@types/lodash": "^4.17.17",
-    "@types/node": "22.15.19",
```
agorer, line 55: Screenshot 2025-06-24 at 08 51 44

It seems that is not a Vercel supported version, we should continue with 22.x.x.

groninge01: ah i didn't check vercel support, reverted that commit

PR #1256, opened 2025-06-26
[LBP] add & update charts

### LINKED ISSUE #1255: LBP chart
add volume/tvl/fees to lbp chart

groninge01:
fixes #1255

### DIFF
--- a/packages/lib/modules/lbp/steps/sale-structure/ProjectedPriceChart.tsx
+++ b/packages/lib/modules/lbp/steps/sale-structure/ProjectedPriceChart.tsx
@@ -17,6 +17,7 @@ type Props = {
   prices: LbpPrice[]
   cutTime?: Date
   isLoading?: boolean
+  gridLeft?: string
 }
 
 export function ProjectedPriceChart({
@@ -26,6 +27,7 @@ export function ProjectedPriceChart({
   prices,
   cutTime,
   isLoading,
+  gridLeft,
 }: Props) {
   const { toCurrency } = useCurrency()
   const theme = useChakraTheme()
@@ -43,7 +45,7 @@ export function ProjectedPriceChart({
 
   const chartInfo: EChartsOption = {
     grid: {
-      left: '5%',
+      left: gridLeft || '10%',
       right: '4%',
       top: '10%',
       bottom: '10%',
--- a/packages/lib/modules/lbp/steps/preview/ProjectedPrice.tsx
+++ b/packages/lib/modules/lbp/steps/preview/ProjectedPrice.tsx
@@ -74,6 +74,7 @@ export function ProjectedPrice({
       <CardBody>
         <ProjectedPriceChart
           endDate={parseISO(endTime)}
+          gridLeft="11%"
           onPriceChange={prices => {
             updateMaxPrice(prices)
             onPriceChange(prices)
--- a/packages/lib/modules/pool/LbpDetail/LbpPoolCharts/LbpPriceChart.tsx
+++ b/packages/lib/modules/pool/LbpDetail/LbpPoolCharts/LbpPriceChart.tsx
@@ -13,6 +13,7 @@ export function LbpPriceChart() {
       <ProjectedPriceChart
         cutTime={now}
         endDate={endTime}
+        gridLeft="7.5%"
         isLoading={isLoading}
         prices={prices}
         startDate={startTime}


packages/lib/modules/lbp/pool/usePriceInfo.ts
```diff
@@ -14,19 +22,29 @@ export type LbpPrice = {
   projectTokenPrice: number
 }
 
+export type HourlyDataPoint = {
```
agorer, line 25: Do we now that is hourly or ti depends on the length of the LBP sale?

packages/lib/modules/lbp/pool/usePriceInfo.ts
```diff
@@ -14,19 +22,29 @@ export type LbpPrice = {
...
+  fees: number
+  volume: number
+}
+
 export function usePriceInfo(chain: GqlChain, poolId: Address) {
   const apiResult = useQuery(GetLbpPriceInfoDocument, {
     variables: {
       poolId: poolId.toLowerCase(),
       chain,
+      dataPoints: 5000,
```
agorer, line 37: Aren't those too many? if the recommended length of an LBP is 5 days and we want hourly snapshots that would be 120, even for a month sale it would be around 720.

packages/lib/modules/lbp/pool/usePriceInfo.ts
```diff
@@ -39,6 +57,37 @@ function toLbpPrices(apiPrices: LbpPriceChartDataFragment[]): LbpPrice[] {
   })
 }
 
+function aggregateToHourlyData(prices: LbpPriceChartDataFragment[]): HourlyDataPoint[] {
+  const hourlyDataMap = new Map<number, HourlyDataPoint>()
+
+  prices.forEach(price => {
+    const timestamp = secondsToMilliseconds(parseInt(price.timestamp.toString()))
```
agorer, line 64: Jarek added `cumulativeFees` and `cumulativeVolume` to the API call (for calculating header stats). We probably don't need this aggregation.

packages/lib/modules/pool/LbpDetail/LbpDetail.tsx
```diff
@@ -10,7 +10,7 @@ import { GqlPoolLiquidityBootstrappingV3 } from '@repo/lib/shared/services/api/g
 import { now } from '@repo/lib/shared/utils/time'
 import { isAfter, isBefore, secondsToMilliseconds } from 'date-fns'
 import { Top10Holdings } from './Top10Holdings'
-import { LbpPoolChartsContainer } from './LbpPoolChartsContainer'
+import { LbpPoolChartsContainer } from './LbpPoolCharts/LbpPoolChartsContainer'
```
agorer, line 13: Why use Kebab / Pascal case for folders? Isn't the kebab case the convention for naming folders in node and most of the frameworks? (it has something to do with OS issues around casing and URL conventions)

packages/lib/shared/utils/chart.helper.ts
```diff
@@ -1,22 +1,81 @@
-export type LabelFormatterParams = {
-  data: any[]
+import { LbpPrice } from '@repo/lib/modules/lbp/pool/usePriceInfo'
```
agorer, line 1: Aren't this functions too specific for LBP to be in a global chart helper file?

packages/lib/modules/lbp/steps/sale-structure/ProjectedPriceChart.tsx
```diff
@@ -27,6 +42,8 @@ export function ProjectedPriceChart({ startDate, endDate, onPriceChange, prices,
 
   const chartInfo: EChartsOption = {
```
agorer, line 43: Screenshot 2025-07-07 at 08 53 59

Some of the changes around here have made the price labels to be cut on the creation page.

agorer:
Screenshot 2025-07-07 at 08 57 23

In a 7 days LBP sale we are showing 11d on the TVL, Volume and Fees charts.

packages/lib/modules/pool/LbpDetail/LbpDetail.tsx
groninge01: both are used so i guess we need to decide what to use only

packages/lib/shared/utils/chart.helper.ts
groninge01: moved them

packages/lib/modules/lbp/steps/sale-structure/ProjectedPriceChart.tsx
groninge01: fixed this

PR #1259, opened 2025-06-26
LBP: fix form validation on page reload

### LINKED ISSUE #1252: Validate LBP creation form on page reload
Image

As we reload info from local storage, we can be in the situation that after reloading the page invalid fields do not show any error. This is due to how the form library computes the `isValid` for a field. It is done based on the form default values but we are changing those to the values locally stored. We must investigate how to set the form values if available without changing the default values.

MattPereira:
Closes #1252


agorer:
https://github.com/user-attachments/assets/17d8ac3d-5248-449e-93d9-745fdf483bf0

Looks like it is still not validating on reload. 😕

groninge01:
when i input the token image url the fields for the website url and x handle are validated also (this also happens on refresh)

https://github.com/user-attachments/assets/ce45d23e-c576-401a-b1bf-142cf7b6c375

agorer:
> when i input the token image url the fields for the website url and x handle are validated also (this also happens on refresh)

Fixed on https://github.com/balancer/frontend-monorepo/pull/1262

MattPereira:
> Looks like it is still not validating on reload. 😕

works on my machine :smiling_face_with_tear: 
https://www.loom.com/share/1606542b4a30482c88d4e28f338f3a0a

packages/lib/modules/pool/LbpDetail/LbpHeader/LbpHeaderTitleDescription.tsx
```diff
@@ -15,7 +15,7 @@ export function LbpHeaderTitleDescription() {
   const socialLinks = [
     {
       iconType: 'x',
-      href: `https://twitter.com/${lbpPool.x}`,
+      href: lbpPool.x ? `https://twitter.com/${lbpPool.x}` : undefined,
```
MattPereira, line 18: @groninge01 

i heard from @agorer that we decided to make twitter handle optional so am suggesting we he icon on pool page if undefined

PR #1264, opened 2025-06-27
[LBP] init using native asset

### LINKED ISSUE #1176: LBP: Allow creation with native asset as collateral
Hook up `wethIsETH` ( add native asset address to lbps.collateralTokens for each supported chain ? )

MattPereira:
Closes #1176 

lmk if u think we should change collateral token symbol/name for the modal summary to be the wrapped native asset

image

### DIFF
--- a/packages/lib/modules/lbp/steps/useCreateLbpStep.tsx
+++ b/packages/lib/modules/lbp/steps/useCreateLbpStep.tsx
@@ -91,38 +91,30 @@ export function useCreateLbpStep(): TransactionStep {
   const { symbol: launchTokenSymbol } = useTokenMetadata(launchTokenAddress, selectedChain)
   const { symbol: reserveTokenSymbol } = useTokenMetadata(reserveTokenAddress, selectedChain)
 
-  const isValidFormStates =
-    saleStructureForm.formState.isValid &&
-    projectInfoForm.formState.isValid &&
-    !!startTime &&
-    !!endTime
-
-  const createPoolInput = isValidFormStates
-    ? {
-        protocolVersion: 3 as const,
-        poolType: PoolType.LiquidityBootstrapping,
-        symbol: `${launchTokenSymbol}-${reserveTokenSymbol}-LBP`,
-        name: `${name} Liquidity Bootstrapping Pool`,
-        swapFeePercentage: parseUnits((fee / 100).toString(), 18),
-        chainId,
-        lbpParams: {
-          owner: owner || userAddress,
-          blockProjectTokenSwapsIn,
-          projectToken: launchTokenAddress as Address,
-          reserveToken: reserveTokenAddress as Address,
-          projectTokenStartWeight: parseUnits(`${projectTokenStartWeight / 100}`, 18),
-          reserveTokenStartWeight: parseUnits(`${reserveTokenStartWeight / 100}`, 18),
-          projectTokenEndWeight: parseUnits(`${projectTokenEndWeight / 100}`, 18),
-          reserveTokenEndWeight: parseUnits(`${reserveTokenEndWeight / 100}`, 18),
-          startTime: BigInt(Math.floor(new Date(startTime).getTime() / 1000)),
-          endTime: BigInt(Math.floor(new Date(endTime).getTime() / 1000)),
-        },
-      }
-    : undefined
+  const createPoolInput = {
+    protocolVersion: 3 as const,
+    poolType: PoolType.LiquidityBootstrapping,
+    symbol: `${launchTokenSymbol}-${reserveTokenSymbol}-LBP`,
+    name: `${name} Liquidity Bootstrapping Pool`,
+    swapFeePercentage: parseUnits('0.01', 18), // TODO: confirm default value?
+    chainId,
+    lbpParams: {
+      owner: owner || userAddress,
+      blockProjectTokenSwapsIn,
+      projectToken: launchTokenAddress as Address,
+      reserveToken: reserveTokenAddress as Address,
+      projectTokenStartWeight: parseUnits(`${projectTokenStartWeight / 100}`, 18),
+      reserveTokenStartWeight: parseUnits(`${reserveTokenStartWeight / 100}`, 18),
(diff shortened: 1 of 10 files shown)


packages/lib/modules/lbp/steps/useCreateLbpStep.tsx
```diff
@@ -77,36 +88,36 @@ export function useCreateLbpStep(): TransactionStep {
...
 
   const createPoolInput = hasRequiredValues
     ? {
         protocolVersion: 3 as const,
         poolType: PoolType.LiquidityBootstrapping,
-        symbol: `${launchTokenSymbol}-${collateralTokenSymbol}-LBP`,
+        symbol: `${launchTokenSymbol}-${reserveTokenSymbol}-LBP`,
         name: `${name} Liquidity Bootstrapping Pool`,
-        swapFeePercentage: parseUnits('0.01', 18), // TODO: confirm default value
+        swapFeePercentage: parseUnits('0.01', 18), // TODO: confirm default value?
```
MattPereira, line 114: did we ever get final decision on default swap fee % ?

packages/lib/modules/tokens/approvals/useTokenApprovalSteps.tsx
```diff
@@ -58,8 +62,16 @@ export function useTokenApprovalSteps({
...
+    () =>
+      approvalAmounts.filter(amount => {
+        // Always filter out native asset
+        if (isSameAddress(amount.address, nativeAssetAddress)) return false
+        // If wethIsEth is true, also filter out wrapped native asset
+        if (wethIsEth && isSameAddress(amount.address, getWrappedNativeAssetAddress(chain)))
+          return false
+        return true
+      }),
+    [approvalAmounts, wethIsEth, chain]
```
MattPereira, line 74: weird this logic didn't already exist since for an add liquidity with native ETH to a pool composed of WETH it shouldnt ask you to approve WETH :thinking: 

if this breaks add liquidity for some reason, there are other ways to prevent an unnecessary wrapped native asset token approval

groninge01: it doesn't ask for approval [here](https://test.balancer.fi/pools/optimism/v2/0x4fd63966879300cafafbb35d157dc5229278ed2300020000000000000000002b) though?

image

groninge01: so summary modal should show unwrapped native asset

packages/lib/modules/lbp/steps/useCreateLbpStep.tsx
```diff
@@ -77,36 +88,36 @@ export function useCreateLbpStep(): TransactionStep {
   const blockProjectTokenSwapsIn = userActions === 'buy_only' ? true : false
 
   const { symbol: launchTokenSymbol } = useTokenMetadata(launchTokenAddress, selectedChain)
-  const { symbol: collateralTokenSymbol } = useTokenMetadata(collateralTokenAddress, selectedChain)
+  const { symbol: reserveTokenSymbol } = useTokenMetadata(reserveTokenAddress, selectedChain)
```
groninge01, line 91: i suggest we do the variable name switch everywhere so it's in sync with the contracts (can use whatever (marketing) name is best for the ui though)

packages/lib/modules/lbp/steps/useCreateLbpStep.tsx
agorer: I think there is going to be some UI to be able to customize it.

agorer: Screenshot 2025-06-30 at 15 15 12

If it's 1% we will have to change the preview though

packages/lib/modules/lbp/steps/useCreateLbpStep.tsx
MattPereira: i strongly agree but @agorer initially pushed back against matching onchain. can't remember the reason tho

packages/lib/modules/lbp/steps/useCreateLbpStep.tsx
```diff
@@ -77,36 +88,36 @@ export function useCreateLbpStep(): TransactionStep {
...
-    userAddress &&
-    launchTokenAddress &&
-    collateralTokenAddress &&
-    projectTokenStartWeight &&
-    reserveTokenStartWeight &&
-    projectTokenEndWeight &&
-    reserveTokenEndWeight &&
-    startTime &&
-    endTime
+    !!launchTokenSymbol &&
```
agorer, line 94: If everything here is taken from sale structure and project forms, can we just check that the forms are valid or is there any reason to check field by field?

packages/lib/modules/lbp/steps/useCreateLbpStep.tsx
groninge01: i would _not_ use the business vocabulary as that is more subject to change than the backend/contracts
also interactions with backend/contracts (on everything) is more frequent than with users (so i think it will be less translating that way instead of more)

packages/lib/modules/lbp/steps/useCreateLbpStep.tsx
agorer: How can you reach the creation modal without filling the `startTime` and `endTime` fields? Even then are they not requiered? (meaning that `saleStructureForm.formState.isValid` will be false?

MattPereira: the error happens instantly as the form renders just after pasting the sale token address before entering any other details

agorer: [issue-1176.patch](https://github.com/user-attachments/files/21039749/issue-1176.patch)

This will more or less solve it (don't know if missed any edge case). You can take a look and use it if you think it works.

MattPereira: thanks for showing me how it is done i like that refactor a lot! :pray:

moving the `LbpCreationProvider` to only wrap the modal fixes so many weird quirks!

PR #1265, opened 2025-06-30
Set create LBP layout for smaller screens

agorer:


packages/lib/modules/lbp/steps/sale-structure/ProjectedPriceChart.tsx
```diff
@@ -27,6 +29,7 @@ export function ProjectedPriceChart({ startDate, endDate, onPriceChange, prices,
     grid: {
       top: '10%',
       bottom: '10%',
+      containLabel: isMobile ? true : false,
```
groninge01, line 32: `isMobile` is already a bool

groninge01 (approved):

PR #1281, opened 2025-07-01
[#1279] Custom fee for LBPs

### LINKED ISSUE #1279: Custom fee for LBPs
Allow to set a custom fee for LBPs with a minimum of 1%.

Image

Image

agorer:
Closes #1279

### DIFF
--- a/packages/lib/modules/lbp/steps/SaleStructureStep.tsx
+++ b/packages/lib/modules/lbp/steps/SaleStructureStep.tsx
@@ -26,11 +26,11 @@ import { InputWithError } from '@repo/lib/shared/components/inputs/InputWithErro
 import { isAddress } from 'viem'
 import { TokenSelectInput } from '../../tokens/TokenSelectInput'
 import { getChainName, getNetworkConfig } from '@repo/lib/config/app.config'
-import { Clipboard, Edit } from 'react-feather'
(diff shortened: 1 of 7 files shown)


packages/lib/modules/lbp/steps/SaleStructureStep.tsx
```diff
@@ -398,6 +404,74 @@ function UserActionsInput({ control }: { control: Control<SaleStructureForm> })
...
+  const [value, setValue] = useState('minimum')
+
+  useEffect(() => {
+    if (feeValue !== 1.0) setValue('custom')
+  }, [feeValue])
+
+  const isBiggerThanMinimum = (fee: number) => {
+    if (fee < 1) return 'LBP swap fees must be set at or above 1.00%'
+    return true
+  }
```
MattPereira, line 427: LBPs inherit from weighted pools so they also have maximum swap fee % 

I tried entering value greater than the max allowed which causes pool create tx to revert

Suggest we consider setting maximum to 10%

https://github.com/balancer/balancer-v3-monorepo/blob/1eec006f7242c3dae41dd502e286e6af7c961e6e/pkg/pool-weighted/contracts/WeightedPool.sol#L50-L51

agorer: done

PR #1282, opened 2025-07-01
LBP: safe wallet support for pool creation

### LINKED ISSUE #1250: Add SAFE multisig support when creating LBPs
We already have SAFE support when extracting the funds as we are using the previous remove liquidity implementation with some cosmetic changes.

MattPereira:
Closes #1250

TODO:
- [ ] figure out how to batch txs for safe
- [ ] figure out if possible to batch create w/ approve & init using `CREATE2` to predict pool addy


packages/lib/modules/lbp/steps/useCreateLbpStep.tsx
```diff
@@ -91,30 +91,33 @@ export function useCreateLbpStep(): TransactionStep {
...
-      projectTokenStartWeight: parseUnits(`${projectTokenStartWeight / 100}`, 18),
-      reserveTokenStartWeight: parseUnits(`${reserveTokenStartWeight / 100}`, 18),
-      projectTokenEndWeight: parseUnits(`${projectTokenEndWeight / 100}`, 18),
-      reserveTokenEndWeight: parseUnits(`${reserveTokenEndWeight / 100}`, 18),
-      startTime: BigInt(Math.floor(new Date(startTime).getTime() / 1000)),
-      endTime: BigInt(Math.floor(new Date(endTime).getTime() / 1000)),
-    },
-  }
+  const createPoolInput =
+    startTime && endTime
```
agorer, line 95: do we still have problems with the times not existing? don't we load this modal only when everything has been filled?

packages/lib/modules/lbp/steps/useCreateLbpSteps.tsx
```diff
@@ -36,20 +44,79 @@ export function useCreateLbpSteps() {
...
-      createLbpStep,
-      ...tokenApprovalSteps,
-      ...(signPermit2Step ? [signPermit2Step] : []),
-      initLbpStep,
-    ],
+    steps: [createLbpStep, ...steps],
   }
 }
 
+function getApprovalAndCreateSteps({
```
agorer, line 79: Isn't this the same / similar function that we have for add liquidity steps? can we somehow generalize them to be shared? (maybe even remove liquidity steps to even though the code is a little different there)

agorer (approved):

packages/lib/modules/lbp/steps/useCreateLbpStep.tsx
MattPereira: yeah was still crashing w/o the time check and wanted to focus on safe wallet stuff so went back to the lazy fix

could spend more time figuring it out in another PR 

image

packages/lib/modules/lbp/steps/useCreateLbpSteps.tsx
MattPereira: yeah thats a good point! will work on the refactor

PR #1335, opened 2025-07-14
feat: make square token images round

groninge01:


agorer (approved):
what is the difference between this approach and using Chakra Circle component?

Issue #1166, opened 2025-06-09
[QA] Compiling a list of smart wallets and testing the main flows with it


agorer:
Regarding issue #1165 , in order to reduce incidents with people not using the same wallets we use in testing. Normally we are using only SAFE wallet (https://app.safe.global/welcome/accounts) to test smart wallet multisig.

agorer:
Smart contract wallets to test:
- SAFE (https://app.safe.global/)
- Fireblocks (https://www.fireblocks.com/platforms/wallets-as-a-service/)
- August digital (https://www.augustdigital.io/)
- Porto (https://www.anchorage.com/platform/self-custody)

PR #1361, opened 2025-07-18
[#1166] show info message when connected with smart contract wallet that is not recognized

agorer:
Screenshot 2025-07-18 at 13 00 00


groninge01 (approved):

packages/lib/modules/web3/wallets/useContractWallet.ts
MattPereira, line 1: so clever :clap:

PR #1362, opened 2025-07-18
LBP update.tsx

0xDanko:
Added a section for LBP and standardized "How Balancer aims to mitigate this risk" on each section


agorer (approved):
Fixed HTML formatting.

PR #1390, opened 2025-07-25
fix: scale back tokens with a rate provider in reclamms

groninge01:
for tokens with a rate provider reclamm uses the underlying to calculate all price values
however in the ui we want to show the rate for the LST token(s) so we need to scale back using the price rate


agorer (approved):
Cannot say I understand what we are calculating though 😆

Issue #56, opened 2024-07-15
Update chart token colors system


garethfuller:

nicolad:
Hi @garethfuller , @uiuxxx , and team 👋

Is this issue still active, or has the color‑mapping system already been addressed elsewhere?  
If it’s still relevant, I’d be happy to take a look and open a PR.

uiuxxx:
Hi @nicolad 

Yes, this issue is still active! 

We would apply these token colors in places like the pool composition table, so that would be a good first place for you to apply the new colors.

<img width="2754" height="1000" alt="Image" src="https://github.com/user-attachments/assets/34adf1a4-7e0b-4b69-bac0-44b5f699c8d9" />

agorer:
After the merge of https://github.com/balancer/frontend-monorepo/pull/1405 we should also use this colors as a loading placeholder for token images on the pool list page.

PR #1405, opened 2025-08-01
[#1393] Speed improvements on first time loading of the app

### LINKED ISSUE #1393: Speed improvements on pool list page
Image

agorer:
Closes #1393 

We used to load tokens and prices on the SSR step but that made the page too big and slow (after an increase on the number of tokens due to not using whitelists anymore). Here we load tokens and prices on the client side in the background after the page has already been rendered.


agorer:
> only suggestion would be shimmering skeleton loading circles instead of the block art for the transition as token images are loading

Pon proposed to use fixed colors for each token as loading placeholders, it should be implemented on https://github.com/balancer/frontend-monorepo/issues/56

PR #1406, opened 2025-08-01
Add New / Trending pools section

uiuxxx:
- Bring back Featured pools module
- With all the recent launches (e.g of HyperEVM and reCLAMMs), interesting new pools should be highlighted.


packages/lib/shared/pages/PoolsPage/PoolsPage.tsx
```diff
@@ -123,6 +134,16 @@ export function PoolsPage({ children, rewardsClaimed24h }: PoolsPageProps) {
           </Suspense>
         </FadeInOnView>
       </DefaultPageContainer>
+      {isBalancer && (
+        <DefaultPageContainer py="0" rounded="2xl">
+          <Box>
+            {!featuredPoolsLoading && featuredPools.length > 0 && (
```
agorer, line 140: The `featuredPools.length > 0` condition could be above next to the `isBalancer` one, if we don't have featured pools we are going to have empty containers.

agorer (approved):

packages/lib/shared/pages/PoolsPage/PoolsPage.tsx
uiuxxx: Good call. Updated.

PR #1408, opened 2025-08-04
[#1385] check if the start time of the LBP is in the past periodically

### LINKED ISSUE #1385: LBP Bug: Start time selector
When creating a new LBP, it is not possible to schedule a start time in the past. However, if you input a start time, go away and then return to complete the LBP after the scheduled start time, no error is displayed. 

For example, I have just returned to this LBP, where the start time is in the past. However, the hint text says 'This sale starts soon'. 

Image

Solution:
- Create some logic on page refresh to check if the start time:
  - If the start time is now in the past
    -  Replace the 'This sale starts soon...' warning message with an error message: `This start date is already in the past. Set it at least 1 hour ahead to ensure the pool is seeded before the sale begins.` 
    - Have the field in an error state, so that the 'Next' button at the bottom is disabled.
  - If in the future, check to see the 'This sale starts soon...' message should be triggered instead.

- Also only show a single message at a time (see below where both are triggered). In this case, just show the error message.

Image

agorer:
Closes #1385

### DIFF
--- a/packages/lib/modules/lbp/steps/SaleStructureStep.tsx
+++ b/packages/lib/modules/lbp/steps/SaleStructureStep.tsx
@@ -59,6 +59,7 @@ import { useUserBalance } from '@repo/lib/shared/hooks/useUserBalance'
 import { LightbulbIcon } from '@repo/lib/shared/components/icons/LightbulbIcon'
 import { now } from '@repo/lib/shared/utils/time'
 import FadeInOnView from '@repo/lib/shared/components/containers/FadeInOnView'
+import { useInterval } from 'usehooks-ts'
 
 export function SaleStructureStep() {
   const { getToken } = useTokens()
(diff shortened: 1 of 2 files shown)


packages/lib/modules/lbp/steps/SaleStructureStep.tsx
```diff
@@ -376,6 +342,111 @@ function LaunchTokenAddressInput({
...
+  value: string
+  saleStart: string
+}) {
+  const validateSaleEnd = (value: string | number) => {
+    if (typeof value !== 'string') return 'End time must be type string'
+    if (!isAfter(parseISO(value), addDays(parseISO(saleStart), 1))) {
+      return 'End time must be at least 24 hours after start time'
+    }
+    return true
+  }
```
MattPereira, line 423: saw you moved similar function `isSaleStartValid` to outside the component context. that is generally best practice because better performance ( but also this case doesnt really matter right? )

PR #1411, opened 2025-08-04
Fix e2e tests

MattPereira:
the account that is being impersonated on fork has no balance of native asset `ETH` so changed to swap `WETH` for `USDC`

also i think `USDC` selector was broken because of recent API changes to token list. 

selecting the little button at the top now

image

### DIFF
--- a/packages/e2e-tests/tests/dev/swap.spec.ts
+++ b/packages/e2e-tests/tests/dev/swap.spec.ts
@@ -1,4 +1,4 @@
-import { impersonate, setForkBalances } from '@/helpers/e2e.helpers'
+import { impersonate } from '@/helpers/e2e.helpers'
 import { expect, test } from '@playwright/test'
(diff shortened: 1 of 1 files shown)


agorer:
> the account that is being impersonated on fork has no balance of native asset `ETH` so changed to swap `WETH` for `USDC`

How can that be possible? don't we use the test accounts created by Anvil? (those have 10000 native token by default)

agorer:
> > How can that be possible? don't we use the test accounts created by Anvil?
> 
> @agorer the account fork is set up to impersonate is not one of the default anvil accounts. the default anvil accounts come with only native `ETH` . the impersonation useful to act as account that has various tokens

Screenshot 2025-08-07 at 08 48 03
Screenshot 2025-08-07 at 08 48 28

It is one of the test accounts, Alberto changed it from the first because people were using them onchain to change permissions to spend and tests became flimsy. So, in theory, it should also have ETH.

PR #1415, opened 2025-08-05
Load token images from pool info if available

agorer:
Instead of waiting to have info for all the tokens we can load images from the pool info that should be already loaded.

### DIFF
--- a/packages/lib/modules/tokens/TokenIconStack.tsx
+++ b/packages/lib/modules/tokens/TokenIconStack.tsx
@@ -40,7 +40,7 @@ export function TokenIconStack({
             borderColor="background.base"
             borderRadius="100%"
             height={`${size + 4}px`}
-            key={tokenAddress}
+            key={tokenAddress + i}
             ml={i > 0 ? getNestingMargin() : 0}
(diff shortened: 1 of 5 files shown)


packages/lib/modules/tokens/TokenIconStack.tsx
```diff
@@ -40,7 +40,7 @@ export function TokenIconStack({
             borderColor="background.base"
             borderRadius="100%"
             height={`${size + 4}px`}
-            key={tokenAddress}
+            key={tokenAddress + i}
```
MattPereira, line 43: assuming you added index because `tokenAddress` could be duped in some situation but im having trouble imagining scenario that would be true. 

could you share example?

MattPereira (approved):

packages/lib/modules/tokens/TokenIconStack.tsx
agorer: It was showing errors on console. The problem can be either due to showing the underlying token on screen and having a couple of tokens with the same one or maybe just the token address being undefined for a small period of time before stabilizing.

PR #1416, opened 2025-08-05
Avoid showing BPT tokens on swap token selection

agorer:


agorer:
We need BPTs for nested pools, so this PR doesn't make sense. If some of them are not supposed to be shown we should do it through the list returned by the API.

PR #1419, opened 2025-08-06
filter validator set for unstaking

franzns:
As a quick workaround to normalize delegations, we shall unstake from a defined set.


apps/beets-frontend-v3/lib/modules/lst/hooks/useGetAmountDelegatedPerValidator.tsx
```diff
@@ -65,10 +65,12 @@ export function useGetAmountDelegatedPerValidator(chain: GqlChain) {
...
-  const amountDelegatedPerValidator = validatorIds.map(validatorId => ({
-    validatorId,
-    amountDelegated: amountResults[validatorId]?.result ?? 0n,
-  }))
+  const amountDelegatedPerValidator = validatorIds
+    .map(validatorId => ({
+      validatorId,
+      amountDelegated: amountResults[validatorId]?.result ?? 0n,
+    }))
+    .filter(validator => ['13', '14', '24', '29', '30'].includes(validator.validatorId))
```
agorer, line 73: would probably add it to some kind of named list so they don't appear as magic numbers.

agorer (approved):

PR #1420, opened 2025-08-06
Allow gas BPT approvals when the wallet does not allow signatures

agorer:
Bug that didn't allow to remove liquidity if signatures where disabled in settings.


packages/lib/modules/pool/actions/add-liquidity/handlers/BoostedUnbalancedAddLiquidityV3.handler.integration.spec.ts
```diff
@@ -14,7 +14,7 @@ describe('When adding unbalanced liquidity for a V3 BOOSTED pool', async () => {
   const handler = selectAddLiquidityHandler(v3Pool) as BoostedUnbalancedAddLiquidityV3Handler
 
   const humanAmountsIn: HumanTokenAmountWithAddress[] = [
-    { humanAmount: '1', tokenAddress: usdcAddress, symbol: 'USDC' },
+    { humanAmount: '0', tokenAddress: usdcAddress, symbol: 'USDC' },
```
MattPereira, line 17: why this change to unbalanced add liq?

PR #1423, opened 2025-08-06
Improve dev on fork instructions

MattPereira:
See the README changes


agorer:
Can we instead of using a make file use docker-compose to launch it? I have been thinking about having the whole Balancer env deployed locally through containers (API + subgraphs) so we are able to test everything (e.g. right now we cannot test the creation of an LBP manually because we need the API to be in sync with the contracts we are using).

MattPereira:
yeah im okay with the docker approach but that will take a while to set up won't it? 

maybe we can remove makefile whenever the docker PR is ready?

PR #1438, opened 2025-08-11
fix: only withdraw top level tokens in recovery mode

### LINKED ISSUE #1437: Can only withdraw wrapped boosted tokens when pool is in recovery mode
Image

vs

Image

groninge01:
fixes #1437


MattPereira (approved):
lgtm

PR #1449, opened 2025-08-12
chore: update packages

groninge01:
- bump date-fns from 2.30.0 to 4.1.0
- bump @tanstack/react-query-devtools from 5.62.11 to 5.84.2
- bump @react-three/drei from 9.117.3 to 10.6.1

### DIFF
--- a/apps/beets-frontend-v3/package.json
+++ b/apps/beets-frontend-v3/package.json
@@ -30,7 +30,7 @@
     "@nikolovlazar/chakra-ui-prose": "^1.2.1",
     "@repo/lib": "workspace:*",
     "@sentry/nextjs": "8.50.0",
-    "date-fns": "^2.30.0",
+    "date-fns": "^4.1.0",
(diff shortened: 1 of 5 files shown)


apps/beets-frontend-v3/package.json
```diff
@@ -30,7 +30,7 @@
     "@nikolovlazar/chakra-ui-prose": "^1.2.1",
     "@repo/lib": "workspace:*",
     "@sentry/nextjs": "8.50.0",
-    "date-fns": "^2.30.0",
+    "date-fns": "^4.1.0",
```
agorer, line 33: This seems like a huge jump, have we checked that no new changes are messing with the app?

agorer (approved):

apps/beets-frontend-v3/package.json
groninge01: that's why `date-fns` recommended running a type check after the upgrade (which caught one error)

PR #1451, opened 2025-08-12
[#1434] change the Safe app link for HyperEVM to Den app

### LINKED ISSUE #1434: Redirect user to Den app instead of Safe app when on HyperEVM
Safe app does not support HyperEVM as of now.

Image

Image

agorer:
Closes #1434 

Also fixes a problem with the Wallet Connect modal that wasn't shown on top of the RainbowKit modal.


packages/lib/modules/web3/safe.hooks.tsx
```diff
@@ -83,3 +86,13 @@ export function useTxHash({ wagmiTxHash }: Props) {
 
   return { txHash, isSafeTxLoading }
 }
+
+export function useSafeAppLink() {
+  const { chain } = useNetworkConfig()
+
+  if (chain === GqlChain.Hyperevm) {
+    return `https://safe.onchainden.com/share/safe-app?appUrl=https%3A%2F%2F${PROJECT_CONFIG.projectId}.fi%2Fpools&chain=hyperevm`
```
groninge01, line 94: because it's hyperevm specific you could just use `balancer` here

groninge01 (approved):

PR #1432, opened 2025-08-08
Warnings for no fiat token price

MattPereira:
Closes #1380

See pool without token price example:
https://mono-test-v3-git-no-price-warning-balancer.vercel.app//pools/hyperevm/v3/0x77d8854b8b555959e890c0d771086517e46a58f0

PR #1452, opened 2025-08-12
fix: no fiat price fixes

groninge01:
a few fixes for #1432

- create map with missing token prices and check that
- add some blank lines for readability
- add some guards to prevent `undefined` in text
- fix hydration errors


MattPereira (approved):
Thanks for cleaning up the mess I made :sweat_smile:

PR #1454, opened 2025-08-12
Avoid showing token icons when they are not available

agorer:
A non existent token appeared on the weekly incentives at pool detail page.

Screenshot 2025-08-12 at 18 32 38

https://balancer.fi/pools/ethereum/v3/0x6c5972311191097d002e804a9bf97c96c54059ed

### DIFF
--- a/packages/lib/modules/pool/PoolDetail/PoolStats/PoolSnapshot/PoolSnapshotValues.tsx
+++ b/packages/lib/modules/pool/PoolDetail/PoolStats/PoolSnapshot/PoolSnapshotValues.tsx
@@ -20,12 +20,17 @@ type PoolStatsValues = {
 }
 
 export function PoolSnapshotValues() {
-  const { chain, pool, tvl } = usePool()
+  const { chain, pool, tvl, isLoading: isLoadingPool } = usePool()
   const { toCurrency } = useCurrency()
 
   const MemoizedMainAprTooltip = memo(MainAprTooltip)
 
-  const { tokens, weeklyRewards, weeklyRewardsByToken } = useGetPoolRewards(pool)
+  const {
+    tokens,
+    weeklyRewards,
+    weeklyRewardsByToken,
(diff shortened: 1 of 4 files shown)


MattPereira (approved):
looks like cause is tokens are initially coming in undefined
image



seems like this solution is sort of band aid rather than fixing root but im okay with that since seems safer than risking knock on effects

groninge01:
suggestion:

- in `PoolSnapshotValues` get `isLoading` from `useGetPoolRewards`
- pass `isLoading` to `TokenIconStack` and `TokenStackPopover`
- add a loading state to both these components

```
  if (isLoading) {
    return (
      <HStack {...rest}>
        {[...Array(2)].map((_, i) => (
          <SkeletonCircle key={`skeleton-${i}`} ml={i > 0 ? -5 : 0} size="24px" zIndex={9 - i} />
        ))}
      </HStack>
    )
  }
```

PS i have a feeling this loading state is required in some other places too where token icons are used

agorer:
> looks like cause is tokens are initially coming in undefined
> 
> seems like this solution is sort of band aid rather than fixing root but also safer than risking knock on effects of changing the rewards hook
> 

I investigated a little bit (as you said it looked suspicious), but the thing here is that we know the rewards but we still don't have info on the tokens. So we know that is going to be two of them, hence the `[undefined, undefined]`, but we don't know which icons. Per Groninge sugestion I have added a loading state for the icons to make this case more visible.

agorer:
> suggestion:
> 
> * in `PoolSnapshotValues` get `isLoading` from `useGetPoolRewards`
> * pass `isLoading` to `TokenIconStack` and `TokenStackPopover`
> * add a loading state to both these components
> 

I have added a loading state to the icon stack component, but left the popover as it was (disabled when the tokens are not available) as having a popover for a loading state seemed a little bit strange.

groninge01 (approved):

PR #1461, opened 2025-08-13
[#1453] allow negative numbers on APR tooltip

### LINKED ISSUE #1453: Allow negative numbers on APR popup
Image

Image

agorer:
Closes #1453 

HyperEVM pools are able to have negative APR numbers.

http://localhost:3000/pools/hyperevm/v3/0x2dc02134ee3b2ee3da099d2854871404f968cf82

### DIFF
--- a/packages/lib/shared/components/tooltips/apr-tooltip/MainAprTooltip.tsx
+++ b/packages/lib/shared/components/tooltips/apr-tooltip/MainAprTooltip.tsx
@@ -16,7 +16,7 @@ import StarsIcon from '../../icons/StarsIcon'
 import { PoolListItem } from '@repo/lib/modules/pool/pool.types'
 import { FeaturedPool } from '@repo/lib/modules/pool/PoolProvider'
 import { Pool } from '@repo/lib/modules/pool/pool.types'
-import { isLBP, isQuantAmmPool } from '@repo/lib/modules/pool/pool.helpers'
+import { isLBP } from '@repo/lib/modules/pool/pool.helpers'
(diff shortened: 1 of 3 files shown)


packages/lib/shared/components/tooltips/apr-tooltip/MainAprTooltip.tsx
```diff
@@ -138,12 +138,15 @@ function MainAprTooltip({
...
+  const aprToShow = apr || getTotalAprLabel(pool.dynamicData.aprItems, vebalBoost, true)
+  const isAprNegative = bn(aprToShow.replace('%', '')).lt(0)
 
   // hoverColor here is used for the text and therefore can use a semantic token
-  const hoverColor = isLBP(pool.type) ? 'inherit' : isAprNegative ? 'font.error' : 'font.highlight'
+  const hoverColor = isLBP(pool.type)
+    ? 'inherit'
+    : isAprNegative
+      ? 'font.warning'
+      : 'font.highlight'
```
MattPereira, line 149: curious why the change from `font.error` to `font.warning` ? kinda seem like warning too close to default tan 

image

MattPereira (approved):

packages/lib/shared/components/tooltips/apr-tooltip/MainAprTooltip.tsx
agorer: It comes from the Slack discussion

Screenshot 2025-08-14 at 10 00 20

The idea is that "In general, I try to keep font.error for situations where the user can't act. In this case, the user can still interact with the pool". The main problem was that we were not showing negative numbers at all though.

PR #1463, opened 2025-08-14
Use normal signatures when the pool is in recovery mode

agorer:
It was not possible to get out of pools like https://beets.fi/pools/sonic/v3/0x5103ea917605463fc497396ba89d6732ce4b2d70 unless you disabled signatures.


MattPereira (approved):
Great work chasing this one down! :muscle:

PR #1480, opened 2025-08-19
[#1307] avoid infinite render loops on batched transactions

### LINKED ISSUE #1307: Bug: batch approval tx w/o using signatures crash
### Summary

happens after a successful add liquidity or pool init when batching approvals but not using signatures

Image

### Error message

```
Error: Maximum update depth exceeded. This can happen when a component repeatedly calls setState inside componentWillUpdate or componentDidUpdate. React limits the number of nested updates to prevent infinite loops.
```

agorer:
Closes #1307


MattPereira (approved):
well done good ser 🔥 

Image

PR #1505, opened 2025-08-28
Build navigational link popover

uiuxxx:
- A new "Build" primary navigation item
- Popover to highlight resources and links for developers

cs 2025-08-28 at 12 23 44@2x

### DIFF
--- a/apps/frontend-v3/lib/components/navs/BuildNavLink.tsx
+++ b/apps/frontend-v3/lib/components/navs/BuildNavLink.tsx
@@ -1,5 +1,3 @@
-'use client'
-
 import { useEffect, useRef } from 'react'
 import {
   Popover,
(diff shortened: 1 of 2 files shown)


apps/frontend-v3/lib/components/navs/BuildNavLink.tsx
```diff
@@ -0,0 +1,84 @@
+'use client'
```
agorer, line 1: Is the `use client` needed?

agorer (approved):

apps/frontend-v3/lib/components/navs/BuildNavLink.tsx
uiuxxx: No! Deleted.

PR #1525, opened 2025-09-03
[#1486] Improve experience when re-locking votes

### LINKED ISSUE #1486: Only submit votes that would benefit from it
In https://github.com/balancer/frontend-monorepo/pull/1325 we changed the vebal voting component to allow resubmitting votes whenever the user has acquired new voting power (through re-locking). There is a particular case here where the user may have voted for some of the gauges after the re-locking and those would not benefit from being submitted again. There is also the possibility that even if we would benefit from the submission we still want to wait for some of them and select which ones we want to vote.

To help with this case we will inform the user about which gauges will benefit from resubmission and also allow the user to select which gauges they want to vote for. For example the previous UI solved this with something like:

Image

Screenshot user address: 0x1888E041D2c9a3DfF6792D03Ddb843c6Dd5508cc

agorer:
Closes #1486

### DIFF
--- a/apps/frontend-v3/lib/vebal/vote/Votes/MyVotes/MyVotes.tsx
+++ b/apps/frontend-v3/lib/vebal/vote/Votes/MyVotes/MyVotes.tsx
@@ -231,18 +231,13 @@ export function MyVotes() {
 
                 {hasNewVotes && (
                   <GridItem colSpan={4} mt="md" w="full">
-                    <Alert status="warning" variant="WideOnDesktop">
+                    <Alert status="warning">
                       <AlertIcon as={AlertTriangle} />
-                      <Stack
-                        alignItems="baseline"
-                        direction={{ base: 'column', lg: 'row' }}
-                        gap={{ base: '0', lg: 'sm' }}
-                      >
+                      <Stack alignItems="baseline" gap="0">
                         <AlertTitle>Resubmit your votes to use your full voting power</AlertTitle>
-                        <AlertDescription>
-                          Votes on pools are set at the time of vote. Since you've added new veBAL
-                          since you original vote, you have addtional voting power that is not being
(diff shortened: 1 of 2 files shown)


apps/frontend-v3/lib/vebal/useVebalUserData.ts
```diff
@@ -22,12 +27,24 @@ export function useVebalUserData() {
     balanceResponse.refetch()
   }
 
+  const snapshots = apiResponse.data?.veBalGetUser.lockSnapshots
+  const lastLockTimestamp =
+    snapshots && snapshots.length > 0
+      ? bn(calculateLastLock(snapshots).timestamp).times(1000).toNumber()
```
groninge01, line 33: you can do `fromUnixTime(calculateLastLock(snapshots).timestamp).getTime()` to remove the magic number

apps/frontend-v3/lib/vebal/vote/Votes/MyVotes/VoteUnderpoweredTooltip.tsx
```diff
@@ -0,0 +1,50 @@
...
+          expiry. You have acquired new veBAL since your last vote. Re-vote to update the gauge with
+          your full veBAL voting power.
+        </Text>
+      </VStack>
+    </PopoverContent>
+  )
+
+  return (
+    <Popover trigger="hover">
+      <>
```
groninge01, line 39: why is the fragment needed here (and elsewhere)?

groninge01 (approved):

apps/frontend-v3/lib/vebal/useVebalUserData.ts
agorer: Have ended up using `secondsToMilliseconds`.

PR #1608, opened 2025-09-24
Make Feedback options more globally available 

uiuxxx:
Currently users can only give feedback at the end of a successful transaction. 

- This PR places feedback options persistently in the header on large screens and within the footer. 
- It also makes a minor typography bug fix for the "Ask questions" link besides the "Give feedback" action in the successful transaction screen.


packages/lib/shared/components/navs/NavBar.tsx
```diff
@@ -166,6 +170,14 @@ export function NavActions({
...
         display: { base: 'none', lg: 'block' },
       },
+      ...(isBalancer
+        ? [
+            {
+              el: <UserFeedback />,
+              display: { base: 'none', lg: 'block' },
+            },
+          ]
+        : []),
```
groninge01, line 180: only show it for Balancer

PR #1612, opened 2025-09-24
Share learn more modal

MattPereira:
- we don't have design / content for pool creation help modal yet
- ~not sure best place to move `LearnMoreModal.tsx`~


groninge01:
> > also there is some specific stuff which could be moved to the project configs
> 
> @groninge01 not sure what should be moved to project configs cus beets docs dont have info about reclamm
> 
> https://docs.beets.fi/beets-dex/pools

not yet no but we can prepare by adding smth like `readMoreUrlLbp` and `readMoreUrlPoolCreator` to the project config
(or do that later)

groninge01 (approved):

PR #1639, opened 2025-10-02
check if relayer approval is needed

### LINKED ISSUE #1638: ContractFunctionExecutionError: The contract function "hasApprovedRelayer" returned no data ("0x").
error from sentry: ContractFunctionExecutionError: The contract function "hasApprovedRelayer" returned no data ("0x").

occurs on chains that do not have a relayer deployed for add/remove liq & swap txns

groninge01:
fixes #1638


MattPereira (commented):
totally unfamiliar with relics / reliquary which seem to be beets specific but could same error from issue happen here?

https://github.com/balancer/frontend-monorepo/blob/032c8793b7ed83822f65cc8e3cc49ff9034e2e31/apps/beets-frontend-v3/lib/modules/reliquary/hooks/useHasApprovedRelayerForAllRelics.tsx#L19

PR #1640, opened 2025-10-02
fix: pool creator reclamm options

groninge01:
make `custom` option a regular radio instead of a checkbox

before:
image

after:
image


MattPereira (approved):
ya this is sneaky big improvement thank you ser 🫡

PR #1663, opened 2025-10-07
Remove the "Create pools" button above the Pools List

uiuxxx:
- Placing the "Create pool" button next to the "Filters" button above the pools list was never the right position for it.
- This PR removes it from there and creates a new builder promo section at the bottom of the page.
- This is in addition to the "Create a pool" links recently added into the "Build" primary navigation link.

create-pool

### DIFF
--- a/packages/lib/modules/pool/PoolList/PoolListFilters.tsx
+++ b/packages/lib/modules/pool/PoolList/PoolListFilters.tsx
@@ -627,18 +627,20 @@ export function PoolListFilters() {
             </PopoverContent>
           </Box>
         </Popover>
-        <Button
-          as={Link}
-          display="flex"
-          gap="2"
-          href={poolCreatorUrl}
-          isExternal
-          ml="ms"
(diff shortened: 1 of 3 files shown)


packages/lib/modules/pool/PoolList/PoolListFilters.tsx
```diff
@@ -643,19 +643,21 @@ export function PoolListFilters() {
...
+            href={poolCreatorUrl}
+            isExternal
+            ml="ms"
+            onClick={() => trackEvent(AnalyticsEvent.ClickPoolListCreatePool)}
+            variant="tertiary"
+          >
+            <Icon as={Plus} boxSize={4} />
+            {!isMobile && 'Create pool'}
+          </Button>
+        )}
```
MattPereira: Image

For Beets, :point_up:  is the only way to get to creator because 

no "Build" popover 

Image


no bottom of page promo 

Image

PR #1695, opened 2025-10-15
Show LBP pool detail on prod

agorer:


groninge01:
can you add `isBalancer` too?

agorer:
> can you add `isBalancer` too?

sure

PR #1701, opened 2025-10-16
feat: add analytics tracking to home page elements

uiuxxx:
- Adds Fathom tracking to navigational elements and items on Pool List page of the app.

### DIFF
--- a/packages/lib/shared/components/navs/NavBar.tsx
+++ b/packages/lib/shared/components/navs/NavBar.tsx
@@ -64,11 +64,7 @@ function NavLinks({
   const { linkColorFor } = useNav()
 
   const handleLinkClick = (analyticsEvent?: string) => {
-    if (
-      isBalancer &&
-      analyticsEvent &&
-      AnalyticsEvent[analyticsEvent as keyof typeof AnalyticsEvent]
-    ) {
+    if (analyticsEvent && AnalyticsEvent[analyticsEvent as keyof typeof AnalyticsEvent]) {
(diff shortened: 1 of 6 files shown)


apps/frontend-v3/lib/components/navs/BuildNavLink.tsx
```diff
@@ -43,6 +44,7 @@ export function BuildNavLink() {
   }, [isOpen, onClose])
 
   const handleToggle = () => {
+    trackEvent(AnalyticsEvent.ClickNavBuild)
```
agorer: The track event code does not seem to differentiate between prod and the dev environments. That could probably skew the results a little.

packages/lib/shared/components/navs/NavBar.tsx
```diff
@@ -61,6 +62,12 @@ function NavLinks({
 }) {
   const { linkColorFor } = useNav()
 
+  const handleLinkClick = (analyticsEvent?: string) => {
```
agorer: Some of the code where we are tracking events is shared with Beets, do we want to track Beets events?

apps/frontend-v3/lib/components/navs/BuildNavLink.tsx
agorer: I see there is some code before that which mades it only run on prod:

```    
// Optional: Only track on production; remove these two lines if you want to track other environments
if (!isProd) return
```

The `zen.balancer.fi` does not point to anything though, we can remove it.

PR #1705, opened 2025-10-16
fix: update lbp create user actions

groninge01:
The first line under `User Rights` should be dynamic.

Setting the action here
image

Before:
image

After: 
image

### DIFF
--- a/packages/lib/modules/lbp/steps/review/OtherSaleDetails.tsx
+++ b/packages/lib/modules/lbp/steps/review/OtherSaleDetails.tsx
@@ -1,4 +1,5 @@
 import { Card, Heading, List, ListItem, Text, VStack } from '@chakra-ui/react'
+import { UserActions } from '@repo/lib/modules/lbp/lbp.types'
 import { fNum } from '@repo/lib/shared/utils/numbers'
 
 export function OtherSaleDetails({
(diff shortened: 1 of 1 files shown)


packages/lib/modules/lbp/lbp.types.ts
```diff
@@ -10,7 +10,7 @@ export type SaleStructureForm = {
   weightAdjustmentType: 'linear_90_10' | 'linear_90_50' | 'custom'
   customStartWeight: number
   customEndWeight: number
-  userActions: 'buy_and_sell' | 'buy_only'
+  userActions: 'buy_and_sell' | 'only_buy'
```
MattPereira: would this change break the radio input?

https://github.com/balancer/frontend-monorepo/blob/f6ab45ff0434f9e81f9151a90af3018a03cea108/packages/lib/modules/lbp/steps/SaleStructureStep.tsx#L533-L534

MattPereira: same here

https://github.com/balancer/frontend-monorepo/blob/f6ab45ff0434f9e81f9151a90af3018a03cea108/packages/lib/modules/lbp/modal/LbpSummary.tsx#L100-L101

groninge01: switched to using enums instead of strings
can you pls check it again?

PR #1708, opened 2025-10-20
fix: relayer is always needed here

groninge01:
for claim (all) and claim & unstake the relayer is always needed

### DIFF
--- a/packages/lib/modules/pool/actions/claim/useClaimAllRewardsSteps.tsx
+++ b/packages/lib/modules/pool/actions/claim/useClaimAllRewardsSteps.tsx
@@ -36,17 +36,14 @@ export function useClaimAllRewardsSteps(params: ClaimAllRewardsStepParams) {
     useClaimAllRewardsStep(params)
 
   const steps = useMemo((): TransactionStep[] => {
-    const steps = [claimAllRewardsStep]
-
-    if (shouldSignRelayerApproval) {
-      steps.unshift(signRelayerStep)
-    } else {
-      steps.unshift(relayerApprovalStep)
-    }
+    const steps: TransactionStep[] = []
 
     if (hasUnclaimedBalRewards) {
-      steps.unshift(minterApprovalStep)
+      steps.push(minterApprovalStep)
     }
+
+    steps.push(shouldSignRelayerApproval ? signRelayerStep : relayerApprovalStep)
+    steps.push(claimAllRewardsStep)
 
     return steps
   }, [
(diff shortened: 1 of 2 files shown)


packages/lib/modules/pool/actions/claim/useClaimAllRewardsSteps.tsx
```diff
@@ -14,12 +12,12 @@ export function useClaimAllRewardsSteps(params: ClaimAllRewardsStepParams) {
     throw new Error('Pools should contain at least one element')
   }
 
-  const relayerMode = useRelayerMode(pool as Pool)
-
   const { chain } = pool
   const chainId = getChainId(pool.chain)
   const hasUnclaimedBalRewards = params.balTokenRewardsQuery.balRewardsData.length > 0
 
+  const relayerMode = 'approveRelayer' // relayer is always needed to claim rewards from gauges
```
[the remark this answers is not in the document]
groninge01: took another look

- removed `pool` prop from `useRelayerMode` where needed (this bypasses the first two checks which are not needed for claim as it's not pool specific)
- added `signRelayerStep`
- check user settings and use signature or gas txn for relayer approval

packages/lib/modules/pool/actions/unstake/useClaimAndUnstakeSteps.tsx
```diff
@@ -28,7 +34,14 @@ export function useClaimAndUnstakeSteps(unstakeParams: UnstakeParams): {
   )
 
   const steps = useMemo((): TransactionStep[] => {
-    const steps = [relayerApprovalStep, claimAndUnstakeStep]
+    const steps = [claimAndUnstakeStep]
+
+    if (shouldSignRelayerApproval) {
+      steps.unshift(signRelayerStep)
```
agorer: doesn't all those unshift look a little difficult to understand?

agorer (approved):

packages/lib/modules/pool/actions/unstake/useClaimAndUnstakeSteps.tsx
groninge01: some steps are conditional so doing it like above isn't the best way to build the array

i can do it first to last if last to first is too difficult?

```
    const steps: TransactionStep[] = []

    if (hasUnclaimedBalRewards) {
      steps.push(minterApprovalStep)
    }

    steps.push(shouldSignRelayerApproval ? signRelayerStep : relayerApprovalStep)
    steps.push(claimAllRewardsStep)

    return steps
```

MattPereira (approved):

PR #1711, opened 2025-10-20
fix lbp claim

mendesfabio:
- the lbp bpt is erc20 token and can be transferred;
- whoever holds the bpt should be able to claim funds and not only the owner set at creation.


joaobrunoah (approved):
LGTM!

PR #1717, opened 2025-10-21
fix: gas estimate speed label

groninge01:
the gas estimate shown in the ui is for the wallet's slowest txn speed option
for example in Rabby that is the `Normal` option and in MetaMask it is the `Low` option

image


groninge01:
> seeing a strange empty white box next to the poppover trigger
> image image

@uiuxxx we need a fallback image for the browser injected wallet

PR #1721, opened 2025-10-21
Better claimable incentives layouts

uiuxxx:
For claimable incentives:
- Order by fiat value
- To make it look more balanced, when there is more than one claimable item, but less than the number of grid cells in the first row, add empty styled cells.

Other minor improvements:
- Improve spacing between warnings on veBAL voting page.
- Add more token colors for popular pools

Screenshots:

Order by fiat value:

cs 2025-10-21 at 16 01 14@2x

More balanced layouts for claimable incentives:

claimable-tokens

More consistent spacing between warnings on veBAL voting page:

cs 2025-10-21 at 16 01 21@2x


groninge01:
> @groninge01 For showing the empty cards to make a more balanced layout, I have also tried improving things for Beets which use a different column system (since there are only two networks). Please check to see that is behaving as expected.

lgtm!

groninge01 (approved):

PR #1724, opened 2025-10-22
fix: add copy button for pool address

### LINKED ISSUE #1723: request: add copy button next to pool address
add copy button next to pool address to make copying the address easier
this functionality is already available from the token popover

Image

groninge01:
fixes #1723


MattPereira (approved):
lgtm

PR #1753, opened 2025-10-29
feat: use path params to init LBP

MattPereira:
### Summary
- refactored to use `LBPCreatePageWrapper` to make easier to share among apps
- path to init looks like `/lbp/create/<chain>/<poolAddress>`
- hide configuration inputs of `saleStructureForm` if pool has already been created
- hydrate lbp config form using on chain reads

https://mono-test-v3-git-flex-init-lbp-balancer.vercel.app/lbp/create/SEPOLIA/0x09ac505a2090ad0b2e8d8361f163ae98a34584e9

image

### Questions
- should we be using search params instead of path params?

### DIFF
--- a/packages/lib/modules/lbp/LbpFormProvider.tsx
+++ b/packages/lib/modules/lbp/LbpFormProvider.tsx
@@ -16,7 +16,7 @@ import { LbpPrice, max, min } from './pool/usePriceInfo'
 import { CustomToken } from '@repo/lib/modules/tokens/token.types'
 import { getNetworkConfig } from '@repo/lib/config/app.config'
 import { getChainId } from '@repo/lib/config/app.config'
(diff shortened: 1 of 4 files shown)


agorer:
> should we be using search params instead of path params?

don't we normally use path params for these things? any advantage of one over the other?

packages/lib/modules/lbp/steps/SaleStructureStep.tsx
```diff
@@ -379,7 +337,7 @@ function SaleStartInput({
   )
 }
 
-function isSaleStartValid(value: string | number) {
+export function isSaleStartValid(value: string | number) {
```
agorer: I we are going to share this functions, they shouldn't be in a component.

PR #1754, opened 2025-10-29
fix: indexeddb error

groninge01:
move entire wagmi config to the client to fix build and sentry errors

image

### DIFF
--- a/packages/lib/modules/web3/WagmiConfig.ts
+++ b/packages/lib/modules/web3/WagmiConfig.ts
@@ -71,7 +71,6 @@ export const wagmiConfig = createConfig({
   chains,
   transports,
   connectors,
-  ssr: true,
 })
 
 /*
@@ -109,7 +108,6 @@ export function impersonateWagmiConfig(impersonationAddress?: Address) {
     chains,
     transports: _transports,
     connectors,
-    ssr: true,
   })
 
   return { connectors, updatedConfig: impersonatedConfig }


packages/lib/modules/web3/WagmiConfig.ts
```diff
@@ -24,44 +24,48 @@ import { isBrowser } from '@repo/lib/shared/utils/app.helper'
...
   https://github.com/rainbow-me/rainbowkit/issues/2232
 */
-if (isBrowser() && !isConnectedToWC()) {
-  const lastConnector = connectors[connectors.length - 1]
-  if (lastConnector({} as any).id !== 'walletConnect') {
-    connectors.push(
-      createWalletConnectConnector({ index: connectors.length, walletConnectProjectId })
-    )
+  if (!isConnectedToWC()) {
+    const lastConnector = connectors[connectors.length - 1]
```
agorer: Is this file executed on the server and the browser? I see that it has the 'use client' directive, but I am not sure if that makes any sense outside components and hooks. If it only executes on the browser, is the conditional above needed? If it executes in both envs, wouldn't `lastConnector` be undefined in the server case and we will have some kind of error in the next line?

groninge01: no just for the wagmi config

```
export const wagmiConfig = createConfig({
  chains,
  transports,
  connectors,
  ssr: true, // <- set to 'false'
})
```

groninge01: apparently but it's only needed on the client

PR #1760, opened 2025-10-30
[#1531] use SDK AddressProvider instead of directly picking the contract addresses for chain conf

### LINKED ISSUE #1531: Use SDK as single source of truth for v3 contract addresses
The SDK has an [AddressProvider](https://github.com/balancer/b-sdk/blob/main/src/entities/inputValidator/utils/addressProvider.ts) class that will throw friendly error if a contract address does not exist on a given chain. 

Note the `AddressProvider` class only supports v3 at the moment but could be expanded to support v2 as well

agorer:
Closes #1531 

Only for V3 contracts for now.


MattPereira (approved):
lgtm

PR #1765, opened 2025-10-31
chore: get hyperevm info from sdk

groninge01:
- bump @balancer/sdk, viem, wagmi, @tanstack/react-query & @rainbow-me/rainbowkit
- import custom info for hyperevm from sdk
- cast wagmiConfig in test as `any` (probably caused by the version bump and it's a pita to type so as it's just test casted to any)

### DIFF
--- a/apps/beets-frontend-v3/app/api/coingecko/chains/route.ts
+++ b/apps/beets-frontend-v3/app/api/coingecko/chains/route.ts
@@ -1,0 +1,5 @@
+import { handleCoingeckoAssetPlatformsRequest } from '@repo/lib/shared/services/coingecko/assetPlatforms'
+
+export async function GET() {
+  return handleCoingeckoAssetPlatformsRequest()
+}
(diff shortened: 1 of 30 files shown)


packages/lib/modules/web3/ChainConfig.tsx
```diff
@@ -86,7 +86,7 @@ const gqlChainToWagmiChainMap: Partial<Record<GqlChainValues, Chain>> = {
   [GqlChain.Mode]: { iconUrl: '/images/chains/MODE.svg', ...mode },
   [GqlChain.Fraxtal]: { iconUrl: '/images/chains/FRAXTAL.svg', ...fraxtal },
   [GqlChain.Sonic]: { iconUrl: '/images/chains/SONIC.svg', ...sonic },
-  [GqlChain.Hyperevm]: { iconUrl: '/images/chains/HYPEREVM.svg', ...hyperEvm },
+  [GqlChain.Hyperevm]: { iconUrl: '/images/chains/HYPEREVM.svg', ...hyperEVM },
```
agorer: Is the format from the SDK the same as the one from viem / wagmi?

agorer: Uff, that is going to be annoying, but it is what it is. Do we know if the SDK updates the dependencies through dependabot?

PR #1770, opened 2025-11-03
V2 exploit warning message

agorer:


agorer:
Looks like it is not going to be used.

PR #1783, opened 2025-11-05
fix: permit2 expiration

### LINKED ISSUE #1784: Permit2 issues
### 1. Expiration is set to wrong date

Expiration is set like this:

`const permitExpiry = get24HoursFromNowInSecs() * 3`

The first part is correct and gets the date/time 24 hours from now. This is for example `1762497175` (`Friday, November 7, 2025 6:32:55 AM`). But when you then do `1762497175 * 3` you get `5287491525` which translates to `Sunday, July 21, 2137 7:38:45 PM`!
<br>

### 2. Allowance will be consumed

So this check will always be `false` after the first approve and subsequent action is done (because `isAllowed` will always be false).

```
const isComplete = () => {
      const isNotExpired = !!expirations && expirations[tokenAddress] > getNowTimestampInSecs()
      const isAllowed = allowanceFor(tokenAddress) >= amountToApprove
      return requiredRawAmount > 0n && isAllowed && isNotExpired
}
```

And any further actions done before expiration (set to 3 days from approve) will consumed more allowance so the check will never become `true`.

groninge01:
fixes #1784

### DIFF
--- a/packages/lib/modules/tokens/approvals/permit2/usePermit2ApprovalSteps.tsx
+++ b/packages/lib/modules/tokens/approvals/permit2/usePermit2ApprovalSteps.tsx
@@ -127,7 +127,8 @@ export function usePermit2ApprovalSteps({
       const isAllowed = allowanceFor(tokenAddress) >= requiredRawAmount
 
       // some expirations were set too far in the future so redo them
-      const hasValidExpiry = !!expirations && expirations[tokenAddress] < permitExpiry // 3 days from now
+      // the expiration shouldn't be more than 3 days from now
+      const hasValidExpiry = !!expirations && expirations[tokenAddress] < permitExpiry
 
       return requiredRawAmount > 0n && isAllowed && isNotExpired && hasValidExpiry
     }


packages/lib/modules/tokens/approvals/permit2/usePermit2ApprovalSteps.tsx
```diff
@@ -120,9 +123,14 @@ export function usePermit2ApprovalSteps({
...
     const isComplete = () => {
-      const isNotExpired = !!expirations && expirations[tokenAddress] > getNowTimestampInSecs()
-      const isAllowed = allowanceFor(tokenAddress) >= amountToApprove
-      return requiredRawAmount > 0n && isAllowed && isNotExpired
+      const isNotExpired = !!expirations && expirations[tokenAddress] > nowInSecs
+      const isAllowed = allowanceFor(tokenAddress) >= requiredRawAmount
+
+      // some expirations were set too far in the future so redo them
+      // the expiration shouldn't be more than 3 days from now
+      const hasValidExpiry = !!expirations && expirations[tokenAddress] < permitExpiry
```
agorer: not sure I understand all of this, but why were some expirations set more than 3 days into the future? 
don't we set the expiration dates?

PR #1794, opened 2025-11-10
fix: add shared Card component for MyTransactions

### LINKED ISSUE #1228: make 'MyTransactions' more composable
make 'MyTransactions', which is used in lbp and pool detail, more composable
an initial setup is [here](https://github.com/balancer/frontend-monorepo/tree/issue-1118-temp)

groninge01:
fixes #1228


groninge01:
> do we want the same column widths here? (left = lbp, right = pool)
> image

leave this for now

PR #1799, opened 2025-11-10
[#1585] Add reasons for blocking flexible/proportional add liquidity

### LINKED ISSUE #1585: Add reasons for blocking flexible/proportional add liq
[version] [pool-type] pools do not support liquidity to be supported proportionally
_Gyro CLP pools do not support liquidity to be supported flexible_

For specific cases:
_This is a nested pool, which does not support liquidity to be supported proportionally_

A general fallback (keep the existing message):
_This pool does not support liquidity to be added proportionally._

agorer:
Closes #1585


packages/lib/modules/pool/actions/LiquidityActionHelpers.spec.ts
```diff
@@ -844,3 +856,91 @@ test('toPoolState keeps pool type when pool is V3 (it does not call mapPoolType)
...
+    }
+
+    expect(requiresProportionalInput(pool)).toBe(true)
+    expect(requiresProportionalInput(pool)).not.toBeUndefined()
+  })
+
+  it('should allow unbalanced add liquidity for other pools (e.g. reClamm)', () => {
+    const pool = getApiPoolMock(sDAIWeighted)
+    pool.type = GqlPoolType.Reclamm
+  })
```
groninge01: test is not finished? also reclamm only allows proportional adds or not?

PR #1805, opened 2025-11-11
chore: bump eslint-plugin-react-hooks

groninge01:
- bump eslint-plugin-react-hooks to 7.0.1
- temporarily disable new/updated rules (these will be fixed in other PRs)

warnings for new rules will be fixed in #1806


agorer (approved):
Can we mark the changes with some kind of FIXME message so we can search for them?

PR #1825, opened 2025-11-14
fix set-state-in-effect linting issues for pool creation

MattPereira:
spent some time learning about this blunder today and wanted some practice :upside_down_face:


groninge01:
this is the rule with the most work i think so tyvm :pray:

PR #1829, opened 2025-11-17
[#1763] Update to Next 16

### LINKED ISSUE #1070: Migrate production builds to turbo
https://areweturboyet.com/

For now, `build --turbo` is still experimental and not production ready so we are only using it in: 
- test preview builds (bal and beets)
- CI dev e2e tests

Once they announce production readiness we should change all builds and merge smoke and dev e2e tests as they both will be able to use turbo builds.

### LINKED ISSUE #1763: Upgrade to next v16
- optionally enable `reactCompiler`: [docs](https://nextjs.org/docs/app/api-reference/config/next-config-js/reactCompiler)
- bump @next/eslint-plugin-next

agorer:
Closes #1763
Closes #1070 

- Removes webpack configuration (was used on production, it should only use Turbopack from now on)
- Activates react compiler

### DIFF
--- a/apps/beets-frontend-v3/next.config.ts
+++ b/apps/beets-frontend-v3/next.config.ts
@@ -2,9 +2,6 @@ import type { NextConfig } from 'next'
 
 /** @type {import('next').NextConfig} */
 const config: NextConfig = {
-  turbopack: {
-    //
-  },
   serverExternalPackages: ['pino-pretty', 'lokijs', 'encoding'],
   logging: {
     fetches: {
--- a/apps/frontend-v3/next.config.ts
+++ b/apps/frontend-v3/next.config.ts
@@ -4,9 +4,6 @@ import type { NextConfig } from 'next'
 
 /** @type {import('next').NextConfig} */
 const nextConfig: NextConfig = {
-  turbopack: {
-    //
-  },
   serverExternalPackages: ['pino-pretty', 'lokijs', 'encoding'],
   logging: {
     fetches: {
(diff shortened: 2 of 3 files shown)


apps/beets-frontend-v3/next.config.ts
```diff
@@ -25,12 +18,21 @@ const config: NextConfig = {
...
+        pathname: '/images/**',
+      },
+    ],
     minimumCacheTTL: 60,
   },
   transpilePackages: ['@repo/lib'],
 
   // Safe App setup
   headers: manifestHeaders,
+  reactCompiler: true,
```
MattPereira: would it be worthwhile / safer to activate with a separate PR after battle testing the react 16 + turbopack change? or maybe [incremental adoption](https://react.dev/learn/react-compiler/incremental-adoption)?

i don't have a strong understanding or opinion here just thoughts / suggestions

apps/beets-frontend-v3/next.config.ts
```diff
@@ -2,17 +2,10 @@ import type { NextConfig } from 'next'
...
-      ...config.resolve.fallback,
-      fs: false,
-      net: false,
-      tls: false,
-      '@react-native-async-storage/async-storage': false, // rainbowkit tries to find this during the build but we're not in react native here
-    }
-    config.externals.push('pino-pretty', 'lokijs', 'encoding')
-    return config
+  turbopack: {
+    //
```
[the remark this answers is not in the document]
groninge01: i think we can just remove it if we're not adding any options to it

apps/beets-frontend-v3/next.config.ts
groninge01: i think the gradual adoption kicks in once we start removing all the manual memoization

PR #1830, opened 2025-11-17
feat: e2e dev tests for beets

groninge01:
add e2e dev tests for beets
note: these are still a bit flaky and therefore need revisiting

### DIFF
--- a/packages/e2e-tests/tests/dev/beets/stake.spec.ts
+++ b/packages/e2e-tests/tests/dev/beets/stake.spec.ts
@@ -13,7 +13,6 @@ test('Stake S to stS on /stake', async ({ page }) => {
   await clickLink(page, 'Stake \\$S')
 
   // The Stake tab should be active by default
-  //await expect(page.getByRole('button', { name: 'S' })).toBeVisible()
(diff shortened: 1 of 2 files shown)


MattPereira:
i think unstake test is failing because input amount of `stS` is greater than amount the fork account owns as a result of the stake test running first. Assuming the e2e tests run in alphabetical order, but not sure.

put up https://github.com/balancer/frontend-monorepo/pull/1885 against this one to test hypothesis

groninge01:
> i think unstake test is failing because input amount of `stS` is greater than amount the fork account owns as a result of the stake test running first. see #1885
> 
> Assuming the tests run in parallel because of config `fullyParallel: true`, the test will still be flaky because stake needs to happen before unstake. I tried setting fork balance for `stS` in `e2e.helpers.ts` but it does not seem to work i think because `stS` is [fancy proxy contract](https://sonicscan.org/address/0xe5da20f15420ad15de0fa650600afc998bbe3955#code) instead of basic erc20

thanks for the pr!
i added a balance for stS at the beginning of the test (and also updated swap so it is less flaky)

PR #1854, opened 2025-11-21
[#1813] eslint rules react-hooks: immutability

### LINKED ISSUE #1813: eslint rules react-hooks: immutability

agorer:
Closes #1813

### DIFF
--- a/packages/lib/shared/hooks/url.hooks.ts
+++ b/packages/lib/shared/hooks/url.hooks.ts
@@ -2,42 +2,26 @@ import { useEffect, useState } from 'react'
 import { isValidUrl, normalizeUrl } from '../utils/urls'
 
 export function useCheckImageUrl(url: string) {
-  const [isChecking, setChecking] = useState(true)
   const [errorMessage, setErrorMessage] = useState<string | undefined>(undefined)
 
   useEffect(() => {
     if (isValidUrl(url) !== true) return
 
     const image = new Image()
+    image.src = normalizeUrl(url)
 
-    image.onload = function () {
-      if (image.width <= 0) {
-        setErrorMessage('Invalid image')
-      }
-
-      setChecking(false)
-    }
-
-    image.onerror = function () {
-      setErrorMessage('Unreachable URL or invalid image')
-      setChecking(false)
-    }
-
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -42,6 +42,7 @@ next-env.d.ts
 # ide
 .vscode
 .idea
+.aider*
```
MattPereira: do you like aider more than claude code?

agorer: Haven't use Claude Code (don't have an account), but I suppose they are more or less the same. Aider allows me to use more than one model / accounts, and has a good integration with the editor I use (emacs). Anyway, don't use AI to write the main parts of the code, just as a butler for the menial tasks. I am sure you have much more experience with AI than me 😂

packages/lib/shared/hooks/url.hooks.ts
```diff
@@ -1,35 +1,39 @@
...
-  }, [image, validUrlError, url])
+      setChecking(false)
+    }
+
+    image.onerror = function () {
+      setErrorMessage('Unreachable URL or invalid image')
+      setChecking(false)
+    }
+
+    // We use an async call to avoid linting issues like
```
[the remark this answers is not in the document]
agorer: Supposedly the rule is there so we don't set states synchronously within an effect because that is bad (as in it could end up in infinite loop), so it probably does not hurt (?)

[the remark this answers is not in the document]
agorer: isn't the microtask queue the underlying tech used for implementing promises? what advantages would it bring here?

packages/lib/shared/hooks/url.hooks.ts
```diff
@@ -1,35 +1,39 @@
...
-    setChecking(true)
-    image.src = normalizeUrl(url)
-  }, [image, validUrlError, url])
+      setChecking(false)
+    }
+
+    image.onerror = function () {
+      setErrorMessage('Unreachable URL or invalid image')
+      setChecking(false)
+    }
```
MattPereira: please test after rebase to see if this still issue 🙏 
- #1878

PR #1874, opened 2025-11-25
fix: functions cannot be passed error in next v16

### LINKED ISSUE #1877: Fix "Functions cannot be passed directly to client components"
happens on a pool details page and probably some other places

Image

page seems to always crashes on local development

Image

sometimes crash on PR preview deployment
https://mono-test-v3-git-fix-usewatch-balancer.vercel.app/pools/sepolia/v3/0x69997AeAadfe4a6691fc376d682b347e1c86C160

Image

groninge01:
fixes #1877 

next 16 is more strict on passing functions/components as props which gives an error for the not-found page

```
Error: Functions cannot be passed directly to Client Components unless you explicitly expose it by marking it with "use server". Or maybe you meant to call this function rather than return it.
```

splitting the `NotFoundPage` component into a server (current component) and client (`NotFoundPageClient`) component solves the error

### DIFF
--- a/packages/lib/shared/pages/NotFoundPage.tsx
+++ b/packages/lib/shared/pages/NotFoundPage.tsx
@@ -1,7 +1,5 @@
-import { DefaultPageContainer } from '@repo/lib/shared/components/containers/DefaultPageContainer'
-import { Button, Heading, VStack, Text } from '@chakra-ui/react'
+import { NotFoundPageClient } from './NotFoundPageClient'
 import { headers } from 'next/headers'
-import Link from 'next/link'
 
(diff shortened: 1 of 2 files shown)


groninge01:
> this happens on pool details page too
> 
>     * [Fix "Functions cannot be passed directly to client components" #1877](https://github.com/balancer/frontend-monorepo/issues/1877)

fyi the not-found-page is always pre-rendered and it was the only place where the error occurred

on the preview url i clicked through all the sepolia pools above the fold and i don't see the error anymore (or any crashes)
can you confirm it?

PR #1875, opened 2025-11-25
Adapt form library to React 19

agorer:


agorer:
@MattPereira can you take a look at this and test the pool creation part? (sorry for the long PR) I will test the LBP part that was also affected.

MattPereira:
Next button seems stuck on disabled for LBP

image

PR #1882, opened 2025-11-26
fix: pool creation token selection filter

### LINKED ISSUE #1883: Fix pool creation token selection filter
we should prevent user from entering invalid token combinations

image

MattPereira:
closes #1883

### Summary
- removes native asset as option for creating pool ( only allow wrapped native asset )
- disables popular tokens that are already selected as pool tokens

image

image


agorer:
Does it make sense to select two times the same token if it is not the native token?

Screenshot 2025-11-28 at 08 18 41

PR #1884, opened 2025-11-28
Update links to new pool creator

MattPereira:
- enable `/create` route on prod for both balancer and beets
- update pool creation links to point to the new one
- leave cow amm pool creator the same at https://pool-creator.balancer.fi/cow


apps/frontend-v3/lib/components/navs/BuildPopover.tsx
```diff
@@ -148,8 +148,7 @@ function CreateAPool() {
             data-group
             display="flex"
             gap="xxs"
-            href="https://pool-creator.balancer.fi/"
-            isExternal
```
groninge01: `<PoolLink>` should only add the upright arrow for external links

groninge01:
> Looks good to me. I'd keep the Beets link in there for now, unless the Beets team strongly disagrees.

there is a beets link on the landing page
so the link to the beets pool creator can be removed here
cc @MattPereira

MattPereira:
@groninge01 removed beets as requested

but it kinda seems to break the symmetry / design a bit :confused: 

image

PR #1889, opened 2025-12-02
Filter out blacklisted tokens for pool creation

### LINKED ISSUE #1888: Pool creation token blacklist
a few gnosis tokens are not compatible with the Balancer v3 vault. see notion for details 👇 

https://www.notion.so/Fix-double-entry-point-issue-with-EURe-25ee395e2281807abb06e8bb7f6ddcbf#25ee395e22818065957fe74eea3ab76a

carry the list over from old pool creator

https://github.com/balancer/pool-creator/blob/main/packages/nextjs/utils/tokenBlacklist.ts

MattPereira:
closes #1888


groninge01:
maybe the list of blocked tokens should be in [metadata](https://github.com/balancer/metadata/blob/main/tokens/blocked.csv)?

MattPereira:
> maybe the list of blocked tokens should be in [metadata](https://github.com/balancer/metadata/blob/main/tokens/blocked.csv)?

we can do that i just didnt bother initially because afaik pool creator is the only use case

Issue #1809, opened 2025-11-11
eslint rules react-hooks: set-state-in-effect


groninge01:
these warnings might be need to fixed in multiple PRs depending on the complexity of the fix

PR #1895, opened 2025-12-03
fix: react-hooks-set-state-in-effect part 1

groninge01:
see #1809


packages/lib/modules/pool/PoolDetail/PoolStats/PoolSnapshot/PoolSnapshot.tsx
```diff
@@ -40,20 +39,16 @@ const TABS = [
 ] as const
 
 export function PoolSnapshot({ ...props }: CardProps) {
-  const [activeTab, setActiveTab] = useState<ButtonGroupOption>(TABS[0])
   const { pool } = usePool()
   const { colorMode } = useColorMode()
+  const defaultTab = hasTotalBalance(pool) ? TABS[1] : TABS[0]
```
agorer: Don't really understand what we are trying to do here. Do you know why the changing of the default balance here depending on if we have some balance or not?

packages/lib/modules/pool/PoolDetail/PoolUserEvents/PoolUserEvents.tsx
```diff
@@ -189,11 +187,12 @@ export default function PoolUserEvents({
     }
   }, [])
 
-  useEffect(() => {
+  const poolEvents: PoolEventItem[] = useMemo(() => {
```
agorer: why do we need the `useMemo` here? shouldn't this be similar to `const poolEvents = !isLoading ? userPoolEvents : []` or even better always return a list from `useUserPoolEvents`?

packages/lib/modules/pool/PoolList/usePoolOrderByState.ts
```diff
@@ -11,19 +10,30 @@ export function usePoolOrderByState() {
   const {
     queryState: { sorting, setSorting, userAddress },
   } = usePoolList()
-  const [orderBy, setOrderBy] = useState(defaultOrderBy)
+
+  const includeUserBalance = !!userAddress
+
+  const orderBy = useMemo(() => {
```
agorer: Same here, why the need for using a memo?

Issue #1818, opened 2025-11-12
Add flow to migrate pools from v2 to v3


agorer:
After the recent exploit on stable v2 pools, there are still some pools with funds in them that should be migrated to similar v3 pools. To help with this task we should show some info on the pool detail page where this is stated and provide the user with a flow that make it easier to do the migration (remove liquidity from the v2 pool and add liquidity to the v3 pool). The mapping of old pools to new ones will be located on the metadata repository (https://github.com/balancer/metadata/tree/main/pools).

PR #1959, opened 2025-12-17
[#1818] show migrate liquidity modal

agorer:
Still having to sort the transaction flow, but this way it is easier to review.

### DIFF
--- a/packages/e2e-tests/tests/dev/balancer/swap.spec.ts
+++ b/packages/e2e-tests/tests/dev/balancer/swap.spec.ts
@@ -10,7 +10,7 @@ test('Swap 1 ETH to USDC)', async ({ page }) => {
   await expect(button(page, 'Connect wallet')).not.toBeVisible()
 
   await clickButton(page, 'ETH')
-  await page.getByText('Wrapped Ether').click()
+  await page.getByText('Wrapped Ether', { exact: true }).click()
 
   await clickButton(page, 'Select token')
   await page.getByText('USDCUSDC').click()


packages/lib/modules/pool/actions/add-liquidity/AddLiquidityProvider.tsx
```diff
@@ -150,9 +150,6 @@ export function useAddLiquidityLogic(urlTxHash?: Hash) {
     enabled,
   })
 
-  /**
-   * Step construction
-   */
```
MattPereira: 😂

packages/lib/modules/pool/migrations/MigrateLiquidityProvider.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  const { getMigration } = usePoolMigrations()
+  const migration = getMigration(protocol, chainId, poolId)
+
+  const { data: oldPoolData } = useFetchPool(chainId, poolId, userAddress)
+  const { data: newPoolData } = useFetchPool(migration?.new.chainId, migration?.new.id, userAddress)
+
+  useEffect(() => {
+    setHumanAmountsIn(amountsOut)
+  }, [amountsOut])
```
MattPereira: how does this avoid upsetting linter about [set-state-in-effect](https://react.dev/reference/eslint-plugin-react-hooks/lints/set-state-in-effect)?

MattPereira:
when you click outside modal it stays open. the only way to get back is "X" button which is different from our normal add / remove modals

image

packages/lib/modules/pool/migrations/PoolMigrationModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  const removeLiquidityReceipt = useRemoveLiquidityReceipt({
+    chain: oldPool?.chain || GqlChain.Mainnet,
+    txHash: removeLiquidityTxHash,
+    userAddress,
+    protocolVersion: oldPool?.protocolVersion as ProtocolVersion,
+    txReceipt: lastTransaction?.result,
+  })
+
+  const isSuccess = !!removeLiquidityTxHash && removeLiquidityReceipt.hasReceipt // FIXME: [JUANJO] just check last tx?
```
MattPereira: ya last tx (add liquidity in this case) is the legacy pattern which seems to work okay

packages/lib/modules/pool/migrations/MigrateLiquidityProvider.tsx
agorer: probably because it is in other hook? (don't really know for sure)

packages/lib/modules/pool/migrations/PoolMigrationModal.tsx
agorer: Seems like it comes from this line `  const quoteAmountsOut = simulationQuery.data?.amountsOut || emptyTokenAmounts(pool)`  on `RemoveLiquidityProvider`. The empty token amounts is what we have here. Not sure what is their purpose and why are they not just an empty array when no info yet (`pool.poolTokens` returns the BPT as well as the pool tokens).

agorer: Changed by using the actionable tokens as the empty state.

agorer:
> maybe center the migrate button?
> 
changed to use the same pattern as the buttons on the debug page
Screenshot 2025-12-18 at 10 57 49

packages/lib/modules/pool/migrations/MigrateLiquiditySummary.tsx
```diff
@@ -0,0 +1,106 @@
...
+      <VStack align="start" spacing="sm">
+        <HStack height="28px" justify="space-between" w="full">
+          <Text fontSize="sm" fontWeight="bold">
+            {title}
+          </Text>
+          <Text color="font.secondary" fontSize="xs">
+            APR
+          </Text>
+        </HStack>
+
```
groninge01: might as well remove this blank line then? ;)

groninge01:
why don't we add `unstake` to the steps? w/ relayer/minter approval if necessary

image

agorer:
> why don't we add `unstake` (when staked on balancer) to the steps? w/ relayer/minter approval if necessary can also batch `withdraw` with it i think

I think @uiuxxx didn't want some huge flow with everything. That is why if the user is staked they have to go first to the unstake flow (whether it is on Balancer or Aura).

PR #1961, opened 2025-12-19
Feat: Add pool token pie chart

uiuxxx:
- Replace the static donut pie chart with a dynamic chart that builds as the user configures pool tokens
- Make the donut pie chart reflect colors of tokens with previously defined colors
- Add tooltips to show token symbols (and weights for weighted pools only). 
- Add some animations for when tokens are added or removed.

cs 2025-12-19 at 16 35 23@2x

### DIFF
--- a/packages/lib/shared/components/imgs/ReClammConfigSvgs.tsx
+++ b/packages/lib/shared/components/imgs/ReClammConfigSvgs.tsx
@@ -66,6 +66,42 @@ function normalizeWeights(tokenCount: number, tokenWeights?: number[]) {
   return weights.map(w => w / total)
 }
 
+function computeArcs(
+  normalizedAddresses: Address[],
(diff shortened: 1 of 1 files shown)


packages/lib/shared/components/imgs/ReClammConfigSvgs.tsx
```diff
@@ -1,376 +1,941 @@
...
+  const cy = 76
+  const outerR = 75
+  const innerR = 45
+  const startAngleBase = -Math.PI / 2
+
+  const midR = (outerR + innerR) / 2
+  const strokeW = outerR - innerR
+  const circumference = 2 * Math.PI * midR
+
+  const arcs = useMemo(() => {
```
agorer: same as before. You can create a function for making the calculations.

PR #1963, opened 2025-12-19
Fix pool creation number input safety

MattPereira:
app crashes if user enters a non numeric character into gyro param input :grimacing: 

image

### Summary 
- use number input for unkown token price and gyro eclp params
- add "suggested value" feature to number input
- simplify error display for number input

### DIFF
--- a/packages/lib/shared/components/inputs/NumberInput.tsx
+++ b/packages/lib/shared/components/inputs/NumberInput.tsx
@@ -106,8 +106,6 @@ export function NumberInput({
                     w="full"
                   >
                     <NumberInputField
-                      max={99}
-                      min={1}
                       pl={isFiatPrice ? '8' : undefined}
                       placeholder={placeholder}
                     />


packages/lib/shared/components/inputs/NumberInput.tsx
```diff
@@ -57,32 +61,56 @@ export function NumberInput({
...
+                        cursor={isSuggestionApplied ? 'default' : 'pointer'}
+                        fontSize="sm"
+                        onClick={isSuggestionApplied ? undefined : onClickSuggestion}
+                        textDecoration={isSuggestionApplied ? 'none' : 'underline dotted 1px'}
+                        textUnderlineOffset="3px"
+                      >
+                        {suggestedValue}
+                      </Text>
+                    </HStack>
+                    {attribution && attribution}
```
agorer: what is the attribution here? does it only make sense with the suggested value?

MattPereira: attribution is for coingecko for price input

image

PR #1979, opened 2025-12-23
Update dependency versions

agorer:


MattPereira (changes requested):
glancing at the deployment failed logs my guess is `viem` version bump changed type requirements for chain definition

MattPereira (approved):

PR #2006, opened 2026-01-02
[#1999] update apollo/client to v4

### LINKED ISSUE #1999: Update apollo/client version to v4
Dependabot recommended to update @apollo/client-integration-nextjs to v0.14.3 but looks like in order to do it we first have to update @apollo/client version to the last major version (at the time of writing v4.0.11). This version has some breaking changes that we will have to have in mind (e.g `useQuery` is imported from a different place).

agorer:
Closes #1999


apps/beets-frontend-v3/app/(app)/pools/page.tsx
```diff
@@ -11,6 +11,8 @@ export default async function PoolsPageWrapper() {
     variables: {},
   })
 
+  if (!stakedSonicData) return null
+
   return (
     <PoolsPage rewardsClaimed24h={stakedSonicData.stsGetGqlStakedSonicData.rewardsClaimed24h}>
```
MattPereira: maybe better UX to just optional chain `?.` here instead of `return null` ?

that way pools page can still render even if something goes wrong with staked fetch ?

PR #2008, opened 2026-01-02
Feat: "build" links set protocol for pool creation form

MattPereira:
### Summary
- replace old create cow amm links
- attatch search params to build popover links to enable:
  - form state update for protocol
  - modal if user in progress for different protocol than the build popover link they selected

image

image

image

image

### DIFF
--- a/packages/lib/modules/pool/actions/create/modal/RestartPoolCreationModal.tsx
+++ b/packages/lib/modules/pool/actions/create/modal/RestartPoolCreationModal.tsx
@@ -21,6 +21,7 @@ import { getPoolTypeLabel } from '@repo/lib/modules/pool/pool.utils'
 import { Address } from 'viem'
 import { PROJECT_CONFIG } from '@repo/lib/config/getProjectConfig'
 import { useProtocolSearchParams } from './useProtocolSearchParams'
+import { useRouter } from 'next/navigation'
 
 interface RestartPoolCreationModalProps {
   modalTitle?: string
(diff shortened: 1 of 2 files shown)


agorer:
Can we make that Cow is selected as the chosen protocol instead of Balancer when clicked?

https://github.com/user-attachments/assets/61c39854-858e-42f7-85ed-bbaabb59e396

agorer:
> > Can we make that Cow is selected as the chosen protocol instead of Balancer when clicked?
> 
> @agorer should we use path or search params for that?
> 
> or maybe other idea?

Not sure how the flow is implemented but I suppose you can have a hash path parameter already? That leave us with the only options of using a search param or some kind of state at some provider (we could use localstorage but that would be ugly).

packages/lib/modules/pool/actions/create/steps/type/ChooseProtocol.tsx
```diff
@@ -20,6 +23,12 @@ export function ChooseProtocol({ control }: { control: Control<PoolCreationForm>
 
     poolCreationForm.reset({ ...INITIAL_POOL_CREATION_FORM, protocol, poolType })
   }
+
+  useEffect(() => {
+    const protocol = searchParams.get('protocol')
+    if (protocol === 'cow') handleChooseProtocol('CoW')
```
MattPereira: @agorer suppose a user started creating a balancer v3 ReClamm but then clicks a "create cow" link

should we reset the form state or let them continue with the protocol / pool type that is already in progress?

Image

agorer: I would say so, but with a warning before just in case it is done by accident.

agorer: It looks like it works from Balancer to Cow but not the other way around.

https://github.com/user-attachments/assets/34bd994b-bc9e-4128-a4e2-78ecf818dbac

PR #2009, opened 2026-01-02
Fix: local dev navbar items display

MattPereira:
this should be fixed permanently now...

BEFORE
image

AFTER
image

explanation per claude code:

```
  In local development:
  - React Strict Mode causes double-mounting
  - Hot Module Reloading can interrupt animations mid-flight
  - Components can remount before the 1s animation completes
  - Result: Elements stuck at opacity: 0 (invisible) but fully rendered and clickable
```

### DIFF
--- a/packages/lib/shared/components/navs/NavBar.tsx
+++ b/packages/lib/shared/components/navs/NavBar.tsx
@@ -321,7 +321,7 @@ export function NavBar({
         <HStack
           animate="show"
           as={motion.div}
-          initial="hidden"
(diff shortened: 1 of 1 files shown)


packages/lib/shared/components/navs/NavBar.tsx
```diff
@@ -321,7 +321,7 @@ export function NavBar({
         <HStack
           animate="show"
           as={motion.div}
-          initial="hidden"
+          initial={process.env.NODE_ENV === 'development' ? false : 'hidden'}
```
agorer: Didn't understand the code before and neither do now 🤷

PR #2022, opened 2026-01-06
Feat: integrate stake dao votemarket incentives

### LINKED ISSUE #1379: Integrate more voting markets for veBAL incentives
- Paladin: https://quest.paladin.vote/#/bal
- Paladin for Aura: https://quest.paladin.vote/#/aura
- StakeDAO: https://votemarket.stakedao.org/balancer
- StakeDAO Campaigns API: https://api-v3.stakedao.org/votemarket/balancer (doesn't have the concept of multiple tokens per campaign but has campaigns divided in multiple periods)

MattPereira:
closes #1379 

### Summary
- fetch campaign incentives from stake dao's vote market API
- filter by campaigns that have `status.voteOpen`
- sum rewards for campaigns with same gauge address 
- enable optimize votes button

image

image

### DIFF
--- a/packages/lib/shared/services/voting-incentives/useVotingIncentives.tsx
+++ b/packages/lib/shared/services/voting-incentives/useVotingIncentives.tsx
@@ -77,7 +77,6 @@ async function getStakeDaoIncentives(): Promise<PoolVotingIncentivesPerWeek[]> {
 }
 
 async function fetchStakeDaoVoteMarket(): Promise<VoteMarketResponse> {
-  console.log('Stake DAO API fetching at', new Date().toISOString())
   const res = await fetch(STAKE_DAO_VOTE_MARKET_URL)
 
   if (!res.ok) throw new Error(`Failed to fetch Stake Dao Votemarket data: ${res.status}`)


apps/frontend-v3/lib/vebal/vote/VoteList/VoteListProvider.tsx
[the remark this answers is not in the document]
agorer: good catch

apps/frontend-v3/lib/vebal/vote/VoteList/VoteListTable/VoteListTableRow.tsx
```diff
@@ -51,8 +51,7 @@ export function VoteListTableRow({ vote, keyValue, ...rest }: Props) {
     ? 'Get veBAL to select and vote on pool gauges'
     : 'Connect your wallet to select and vote on pool gauges.'
 
-  const missingDataReason =
-    'Bribe data is temporarily unavailable during the transition from Hidden Hand to Votemarket'
+  const missingDataReason = 'There is currently no bribe data for this pool'
```
groninge01: maybe use `incentive` instead of `bribe` (everywhere i guess)?

MattPereira:
> why do some of the numbers in both images (Stake Dao page and Balancer) differ? is because they are taken at different times? some kind of caching?

i must have just taken OG screen shots some time apart and market prices changed.

 i though it best to use token price from stake dao API in case our API doesnt have price for some future incentive token

image

apps/frontend-v3/lib/vebal/vote/VoteList/VoteListTable/VoteListTableRow.tsx
MattPereira: less fun but is prolly better for legal liability risks :joy:

MattPereira:
> what does it mean that isBlacklisted for a campaign? we had a blacklisting mechanism that allowed us to remove Aura votes (they had a different set of incentives). Does that still need to be done?

@agorer im still working on understanding the old blacklist mechanism [here](https://github.com/balancer/frontend-monorepo/blob/integrate-stake-dao/apps/frontend-v3/lib/vebal/vote/Votes/MyVotes/incentivesBlacklist.ts) and if `vlAURA` has separate rewards from regular `veBAL`

PR #2026, opened 2026-01-08
Issues with min deposit errors

agorer:
- Avoid showing scientific notation on screen (e.g. 1e-7)
- After talking with Juani, the SC's rounds here and there (by adding or removing some wei) so we are not able to do the exact calculation that they do. Added 1% buffer to all the mins to avoid errors on small amounts.


agorer:
Ended up rounding the mins to the number of decimals of the token.

PR #2027, opened 2026-01-09
[#2000] update to sdk v5

### LINKED ISSUE #2000: Update @balancer/sdk to v5.0.0
This update implies updating viem to v2.42.0 but that would also mean to update wagmi to the new major version (at the time of writing v3.1.3) which have breaking changes (e.g. useBalance no longer supports asking for any token balance and is only used for the native token balance).

- https://wagmi.sh/core/guides/migrate-from-v2-to-v3#migrate-from-v2-to-v3
- https://wagmi.sh/core/guides/migrate-from-v1-to-v2#deprecations

agorer:
Closes #2000 

Also updated viem and wagmi to the last versions.


groninge01:
optionally could also update rainbowkit

PR #2044, opened 2026-01-13
feat: add popover to unstake too

### LINKED ISSUE #1817: add tooltip to 'Unstake' button
Image

_At minimum, I think we should do the “disabled button tooltip” pattern that we use elsewhere on the site.
i.e. For people with a staked position on Aura (with nothing staked on Balancer), with should show a tooltip on the disabled button of “Aura staking / unstaking is not supported on the Balancer UI. Go to Aura to manage this.”_

_We should probably also add a state for the edge case where someone is staking on both Balancer and Aura. In this case, we could show an enabled ‘Unstake’ button with the upwards caret, and do a similar interaction to the Stake button with caret_

groninge01:
fixes #1817

aura only staked
image

aura & balancer staked
image

balancer only staked
image

### DIFF
--- a/packages/lib/modules/pool/PoolDetail/StakeUnstakeButton.tsx
+++ b/packages/lib/modules/pool/PoolDetail/StakeUnstakeButton.tsx
@@ -110,8 +110,8 @@ export function StakeUnstakeButton({ pool, action }: StakeUnstakeButtonProps) {
   const auraStakedBalanceUsdFormatted = toCurrency(calcAuraStakedBalanceUsd(pool))
 
   const hidePopover = !getNetworkConfig(pool.chain).hasAura
-  const stakeOnBalancer = () => router.push(`${pathname}/stake`)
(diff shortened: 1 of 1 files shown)


packages/lib/modules/pool/PoolDetail/StakeUnstakeButton.tsx
```diff
@@ -0,0 +1,221 @@
...
+            {action === 'stake' && (apr ? `${fNum('apr', apr)} APR` : 'N/A')}
+            {action === 'unstake' && usdValue}
+          </Text>
+        </VStack>
+      </HStack>
+    </Button>
+  )
+}
+
+export function StakeUnstakeButton({ pool, action }: StakeUnstakeButtonProps) {
```
agorer: Not sure I would have one component for both buttons even if the share some structural properties. The problem I have here is that we just have a lot conditional code.

PR #2045, opened 2026-01-13
chore: api cleanup

groninge01:
a lot of deprecated/legacy code is removed in the api
this PR syncs the frontend to those changes

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -5,7 +5,7 @@ on:
     types: [opened, synchronize]
 
 env:
-  NEXT_PUBLIC_BALANCER_API_URL: https://api-v3.balancer.fi/graphql
+  NEXT_PUBLIC_BALANCER_API_URL: https://test-api-v3.balancer.fi/graphql
   NEXT_PUBLIC_WALLET_CONNECT_ID: ${{ secrets.NEXT_PUBLIC_WALLET_CONNECT_ID }}
   NEXT_PRIVATE_ALCHEMY_KEY: ${{ secrets.PRIVATE_ALCHEMY_KEY }}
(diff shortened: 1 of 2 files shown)


packages/lib/modules/pool/pool.helpers.ts
```diff
@@ -235,19 +234,18 @@ export function getPoolHelpers(pool: Pool, chain: GqlChain) {
 
 export function hasNestedPools(pool: Pool) {
   if (!pool.poolTokens) return false
-  // The following discriminator is needed because not all pools in GqlPoolQuery do have nestingType property
-  // and the real TS discriminator is __typename which we don't want to use
```
agorer: do you know why we don't want to use `__typename` if we are already using it in some other places like `isComposableStablePool`?

packages/lib/modules/tokens/__mocks__/token.builders.ts
```diff
@@ -91,23 +89,11 @@ export function someGqlTokenMocks(symbols: string[]): GqlPoolTokenDetail[] {
...
 
-export function aTokenExpandedMock(
-  ...options: Partial<GqlPoolTokenExpanded>[]
-): GqlPoolTokenExpanded {
+export function aTokenExpandedMock(...options: Partial<GqlPoolTokenDetail>[]): GqlPoolTokenDetail {
   const symbol = (options[0]?.symbol as FakeTokenSymbol) || 'BAL'
   const defaultToken: TokenBase = fakeTokenBySymbol(symbol)
   return Object.assign({}, defaultToken, ...options)
 }
-export function aTokenExpandedBySymbolMock(
```
agorer: unused?

PR #2049, opened 2026-01-14
fix: show/hide hh alert

groninge01:
- only show hh alert for hh claims
- fix showing placeholders when no network claims
- fix types


packages/lib/modules/portfolio/PortfolioClaim/ClaimNetworkPools/ClaimNetworkPools.tsx
```diff
@@ -290,7 +292,7 @@ export function ClaimNetworkPools() {
                             textTransform="capitalize"
                             tooltip="Available until June 30, 2026"
                           />
-                        )
+                        ) as ReactNode
```
agorer: isn't this component always a ReactNode?

groninge01: it should be
issue was in the child component -> fixed and removed assertion here

Issue #2037, opened 2026-01-12
Feat: improve navigation UX for LBP and pool creation


MattPereira:
Pon requested we modify path for each form step

e.g.
/create/step-1-type
/create/step-2-tokens
/create/step-3-details
/create/step-4-fund

The main reason is to get browser history to be more predictable. For example, sometimes I press the back button after completing several steps, and it takes me back to the last page I was on (instead of the last step).

PR #2068, opened 2026-01-19
Feat: update path for lbp form steps

### LINKED ISSUE #2067: handle LBP page paths

MattPereira:
closes #2067 & #2037

- shared `useFormSteps` hook for pool and LBP creation to handle URL & step state helpers
- simplify token icon url validation with async function fed to RHF `validate`
- tests for LBP and create pool pages
  - run through all steps to the end
  - reset form for each step

### DIFF
--- a/packages/e2e-tests/tests/dev/balancer/create-lbp.spec.ts
+++ b/packages/e2e-tests/tests/dev/balancer/create-lbp.spec.ts
@@ -22,14 +22,21 @@ async function doSaleStructureStep(page: Page, { continue: shouldContinue = fals
   await dateInputs.last().fill(toISOString(Date.now() + oneWeekInMs).slice(0, 16))
 
   await expect(page.getByRole('heading', { name: 'Seed initial pool liquidity' })).toBeVisible()
+
+  const nextButton = button(page, 'Next')
+
+  await expect(nextButton).toBeDisabled()
   await page.getByLabel('Sale token').fill('100')
   await page.getByLabel('Collateral token').fill('1')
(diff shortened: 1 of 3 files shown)


packages/lib/modules/lbp/LbpFormAction.tsx
```diff
@@ -13,21 +13,19 @@ import { useFormState, useWatch } from 'react-hook-form'
...
-    projectInfoForm,
     poolAddress,
   } = useLbpForm()
   const selectedChain = useWatch({ control: saleStructureForm.control, name: 'selectedChain' })
   const previewModalDisclosure = useDisclosure()
   const nextBtn = useRef(null)
   const { copyToClipboard, isCopied } = useCopyToClipboard()
   const saleFormState = useFormState({ control: saleStructureForm.control })
-  const projectInfoFormState = useFormState({ control: projectInfoForm.control })
-  const isFormStateValid = saleFormState.isValid && projectInfoFormState.isValid
```
MattPereira: the `projectInfoFormState.isValid` was causing problems because it would return `false` with no errors which caused modal not to render. spent a while trying to understand why by reviewing & refactoring the async token icon url validation but still stumped 

i think we are okay removing the check because user should not be able to reach final step where modal is rendered without valid values for project info step inputs

MattPereira:
> I did not check if this is also broken on main but next button is active while the checkbox is unchecked on step 2
After clicking it the UI goes to step 3

@groninge01 thanks for finding. have unable to recreate myself but will add test coverage 

image

PR #2110, opened 2026-02-04
fix: voting incentives filter

MattPereira:
### Summary
- remove rewards that are for vlAURA only 

from the stake DAO team:

> you need to filter out the campaign with a whitelist (isBlacklist : false) and some addresses. There are only for vlAURA, or maybe some other market (maybe in the future)
> For the ones with vlAURA as blacklist, you can account everything
> 
> For the ones with no blacklist (addresses empty), there is no clear allocation between veBAL and vlAURA as said, it's who votes takes it, so i guess you can show the total amount too (that's what we are doing)

time of screenshots is ~19:00 UTC on Feb 4 

### Before
- 16 total campaigns which does not match how votemarket shows veBAL rewards

image

### After:
- 12 total campaigns matches votemarket 

image


MattPereira:
>  have to really understand all the possible types of campaigns so we can calculate the right number of votes. Do we have examples of all the possible types (only aura, only veBAL, mixed)?

- #2115

PR #2130, opened 2026-02-11
fix: extend add/remove liq handlers

### LINKED ISSUE #2077: extend `AddLiquidityHandler` and `RemoveLiquidityHandler`
so the handler stays typed when used in `RelicAddLiquidityProvider` and the guards used below are not needed anymore

https://github.com/balancer/frontend-monorepo/blob/a49c9b57edcc3cc5154a6038e00fff8fb3fbc363/apps/beets-frontend-v3/lib/modules/reliquary/hooks/useReliquaryAddLiquidityStep.tsx#L55-L58

groninge01:
fixes #2077

checked add/remove liq for reliquary and regular pool

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -12,6 +12,7 @@ env:
   TEST_ACCOUNT_MNEMONIC: ${{ secrets.TEST_ACCOUNT_MNEMONIC }}
   TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
   TURBO_TEAM: ${{ vars.TURBO_TEAM }}
+  SLACK_BOT_TOKEN: ${{secrets.SLACK_BOT_TOKEN}}
 
 jobs:
   # Skip this job as it is already part of the E2E-Smoke-Test job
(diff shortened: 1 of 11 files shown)


packages/lib/modules/pool/actions/add-liquidity/AddLiquidityProvider.tsx
```diff
@@ -38,6 +41,15 @@ import { ApiToken } from '@repo/lib/modules/tokens/token.types'
...
 
+type AddLiquidityHandlerSelector<THandler extends AddLiquidityHandler = AddLiquidityHandler> = (
+  pool: Pool,
+  wantsProportional: boolean
+) => THandler
+
+type AddLiquidityStepsHook<THandler extends AddLiquidityHandler = AddLiquidityHandler> = (
+  params: AddLiquidityStepsParams<THandler>
+) => ReturnType<typeof useAddLiquidityStepsBase>
+
```
MattPereira: Can you help me understand the preference for using the THandler generic on the providers?

The reliquary handlers already implement the base handler interfaces, and the actual customization mechanism is the callback props (handlerSelector, useRemoveLiquiditySteps), which work the same with or without the generic.

[Example commit ](https://github.com/balancer/frontend-monorepo/commit/af804e08291f66c3c5f95aed96e283640dcc9e8e)that removes the generic and replaces the handler union types in the reliquary hooks with the base interface type

groninge01: closing this PR then
@MattPereira can you create a PR for your branch?

PR #2152, opened 2026-02-17
chore: setup renovate config

MattPereira:
- install app at https://github.com/apps/renovate
- dependabot security alerts should still be active because separate GH feature controlled by repo settings

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": ["config:recommended"],
+  "extends": ["config:best-practices"],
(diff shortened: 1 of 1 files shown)


MattPereira:
> but, with dependabot we had the same alerts as the SDK (which we have to sync viem releases). Is the SDK also changing to this one?

not sure i understand what you mean — we still have to manually match viem to SDK during version bumps regardless of which bot we use.

my motivation for this was to avoid having to manually close a bunch of PRs, do the search replace alls, and then run `pnpm install` to change the lockfiles

With renovate bot the bump PRs should be sitting there ready to merge if all the CI passes

PR #2154, opened 2026-02-18
chore: unskip remove liquidity tests

### LINKED ISSUE #2133: Unskip remove liquidity tests

MattPereira:
closes #2133


packages/lib/modules/pool/actions/remove-liquidity/useRemoveLiquidityStep.integration.spec.tsx
```diff
@@ -1,57 +0,0 @@
...
-//       }
-//     },
-//     {
-//       wrapper: Providers,
-//     }
-//   )
-//   return result
-// }
-
-test.skip('Throws error when user tries to remove liquidity in a pool where they does not have balance', async () => {
```
MattPereira: the UI blocks user from attempting this even if they cheat the path

Image

agorer: checking that would be a good test then, but that should be some kind of component test.

PR #2158, opened 2026-02-19
Joinable pools filter

uiuxxx:


packages/lib/modules/pool/PoolList/PoolListProvider.tsx
```diff
@@ -62,13 +286,14 @@ export function usePoolListLogic({
   }, [userAddress])
 
   return {
-    pools: poolsData,
-    count: data?.count || previousData?.count,
+    pools: filteredPools,
+    count: joinablePools ? filteredPools.length : data?.count || previousData?.count,
     queryState,
-    loading,
+    loading: loading || (joinablePools && isJoinableBalanceLoading),
```
agorer: isJoinableBalanceLoading already has joinablePools as conditions, this seems like a duplicate.

agorer (approved):

PR #2177, opened 2026-02-23
fix: default swap handler integration test

MattPereira:
`result.hopCount` fluctuates because we have mainnet fork running on the latest block so as liquidity pool conditions change the route the SOR chooses changes

fixing mainnet fork to specific block number should make tests less flaky 

but we will have to remember to bump the mainnet fork block number when adding tests for contracts that were not deployed prior to the chosen block number ( which we already have to do for all forks that are not mainnet )

also anytime fork block number gets bumped a handful of tests may fail as state of blockchain has changed which would require updating tests or setting looser expectations like `expect(result.hopCount).toBeGreaterThan(0)`

wdyt is best approach?

### DIFF
--- a/packages/test/anvil/anvil-setup.ts
+++ b/packages/test/anvil/anvil-setup.ts
@@ -54,9 +54,8 @@ const ANVIL_PORTS: Record<ChainIdWithFork, number> = {
 }
 
 /*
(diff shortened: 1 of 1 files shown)


groninge01:
reset the fork to a specific block # for each test separately?

groninge01 (approved):

PR #2186, opened 2026-02-26
[#18] Swap route visualization

### LINKED ISSUE #18: Spike swap route visualization
Use sankey diagrams from echarts to represent swap routes.

agorer:
Closes #18 

I have been testing using GHO -> USDC and BAL -> USDC but it is difficult to test all the cases as paths can change with each query.


agorer:
> should the vertical row split always be 50/50 or reason for thick bottom row?
> 
> 6 hopper gonna cost a lotta gas 😂
> 
The height is related to the share of the input token that is being swapped through that path. I have only seen 50-50 and 75-25 splits.

PR #2210, opened 2026-03-05
feat: option to boost underlying token during pool creation

### LINKED ISSUE #1510: [Pool Creation] : option to boost underlying token
Will require we swap underlying for boosted OR deposit underlying to protocol vault

Also consider how that will change amounts for fund step ( after swap or deposit user will have less than amount entered of underlying ) 

Image

MattPereira:
closes #1510

Note that the reclamm pool factory is still disabled at the smart contract level

### Tokens Step

warning if user selects underlying that has corresponding erc4626 from our whitelist

image

image

### Fund Step
user inputs wrapped amounts which are used to `previewMint` on the erc4626 to get underlying amount required

image

### Transactions Steps
`onSuccess` of deposit underlying, update the init amounts to match how much wrapped user actually received

image


packages/lib/modules/tokens/approvals/useTokenApprovalSteps.tsx
```diff
@@ -161,14 +161,16 @@ export function useTokenApprovalSteps({
       return errors
     }
 
-    const isTxEnabled = !!spenderAddress && !tokenAllowances.isAllowancesLoading
+    const resolvedSpender =
+      typeof spenderAddress === 'function' ? spenderAddress(tokenAddress) : spenderAddress
```
groninge01: you created a function for this in `useTokenAllowances`

PR #2211, opened 2026-03-05
UI/swap route viz

uiuxxx:
Updated the Swap Routes visual to use a popover instead of to display inline within the accordion. 

routes

### DIFF
--- a/packages/lib/modules/swap/RoutesCard.tsx
+++ b/packages/lib/modules/swap/RoutesCard.tsx
@@ -24,11 +24,11 @@ import { ArrowRight } from 'react-feather'
 import { useCurrency } from '@repo/lib/shared/hooks/useCurrency'
 import { getCompositionTokens } from '../pool/pool-tokens.utils'
 import { PoolToken } from '../pool/pool.types'
-import React from 'react'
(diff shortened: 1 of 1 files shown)


packages/lib/modules/swap/RoutesCard.tsx
```diff
@@ -56,88 +59,158 @@ export function RoutesCard({ paths, chain, totalInputAmount, totalOutputAmount }
...
+        overflow="hidden"
+        p="0"
+        rounded="lg"
+        w={{ base: '350px', md: 'max-content' }}
+      >
+        <VStack p="4" shadow="2xl" spacing="3">
+          <VStack align="start" w="full">
+            <Text color="font.primary" fontSize="md" fontWeight="bold" pb="1">
+              Proposed: {paths && paths.length > 1 ? `${paths.length} paths, ` : ''}
+              {maxHops} hops via Balancer v{protocolVersion}
```
agorer: Given that we also show this on Beets (if I am not mistaken) should we have references to Balancer? (even if they are to the protocol)

agorer (approved):

PR #2212, opened 2026-03-05
[#234] use user locale decimal separator on token input

### LINKED ISSUE #234: Consistent commas/dots in token inputs

agorer:
Closes #234 

- Both comma and period keys can be used to input the decimal separator
- On input if not the separator defined by the user locale it is replaced by the right one

### DIFF
--- a/packages/e2e-tests/tests/dev/balancer/swap.spec.ts
+++ b/packages/e2e-tests/tests/dev/balancer/swap.spec.ts
@@ -8,7 +8,7 @@ test('Swap 1 ETH to GHO)', async ({ page }) => {
 
   await impersonate(page, defaultAnvilAccount)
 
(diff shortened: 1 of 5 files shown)


agorer:
> it keeps defaulting to `en_US` which is not my browser locale shouldn't the locale be retrieved from `navigator.language` first?

I have tried both en_US and es_ES and `new Intl.NumberFormat()` seems to be picking the right locale. 

Screenshot 2026-03-06 at 09 52 36
Screenshot 2026-03-06 at 09 53 03

agorer:
The tests are fixed. In the end, after talking with @uiuxxx I have only changed the token input so it replaces the comma with periods. That way both can be used to input the number but it would stay format consistent with the rest of the numbers around the page.

Issue #2236, opened 2026-03-13
Seedless LBP


agorer:
https://www.figma.com/design/Njrn6DnKSQYzl92Yzx2FEV/v3-app-LBP?node-id=2449-18704&p=f&t=YQJH9djw37moUxDx-0

PR #2255, opened 2026-03-20
[#2236] allow creation of seedless LBPs

agorer:
Still missing pool detail updates and some API changes.


groninge01:
image

radio button should be aligned with yes/no
also space between choice and sub text should be smaller?

PR #2343, opened 2026-04-15
fix: joinable pools rpc timeouts

groninge01:
checking joinable pools fires a lot of rpc calls to fetch all the users tokens

on localhost all rpc calls are ok but on prod they are running into timeouts

this change splits the rpc calls in batches to prevent timeouts

also the rpc fallback urls are updated as some were no longer available


packages/lib/modules/tokens/useWalletTokenBalances.ts
```diff
@@ -13,11 +13,21 @@ import { useUserAccount } from '../web3/UserAccountProvider'
...
-import { useMemo } from 'react'
+import { useMemo, useCallback } from 'react'
 import { captureNonFatalError } from '@repo/lib/shared/utils/query-errors'
 
 const MIN_TOKEN_VALUE_USD = 1
 const BALANCE_STALE_TIME = 30_000
+const PARALLEL_CHAINS = 3
+
+// Split array into chunks of specified size
+function chunkArray<T>(array: T[], size: number): T[][] {
```
agorer: this could probably go to the array utils file.

packages/lib/modules/tokens/useWalletTokenBalances.ts
```diff
@@ -123,15 +146,19 @@ export function useWalletTokenBalances(chains: GqlChain[], enabled: boolean) {
...
-    })
+    }
 
     return result
-  }, [enabled, chains, balanceQueries, priceFor])
+  }, [enabled, chains, balanceQuery.data, priceFor])
 
   const isLoading =
-    isLoadingTokens || isLoadingTokenPrices || balanceQueries.some(q => q.isLoading || q.isFetching)
+    isLoadingTokens || isLoadingTokenPrices || balanceQuery.isLoading || balanceQuery.isFetching
```
agorer: why do we need the isFetching? isn't it synched with the isLoading?

packages/lib/modules/web3/ChainConfig.tsx
```diff
@@ -32,20 +32,20 @@ import { hyperEVM } from '@balancer/sdk'
   https://viem.sh/docs/clients/transports/fallback#fallback-transport
 */
 export const rpcFallbacks: Partial<Record<GqlChainValues, string | undefined>> = {
-  [GqlChain.Mainnet]: 'https://eth.llamarpc.com',
-  [GqlChain.Arbitrum]: 'https://arbitrum.llamarpc.com',
-  [GqlChain.Base]: 'https://base.llamarpc.com',
-  [GqlChain.Avalanche]: 'https://avalanche.drpc.org',
+  [GqlChain.Mainnet]: 'https://1rpc.io/eth',
```
agorer, line 35: are we hitting these fallbacks in production?

agorer (approved):

packages/lib/modules/web3/ChainConfig.tsx
groninge01: once drpc cancels we do

packages/lib/modules/tokens/useWalletTokenBalances.ts
groninge01: moved!

packages/lib/modules/tokens/useWalletTokenBalances.ts
groninge01: removed!

PR #2366, opened 2026-04-20
fix: token balances integration test

groninge01:
`TokenBalancesProvider` assumes all requested tokens belong to the same chain, but this integration test was passing a mixed-chain slice from `allFakeGqlTokens`. 

That made the balance query flaky in CI because some token reads were executed against the wrong chain, so the test now filters to a single chain and asserts against that deterministic set.


MattPereira (approved):
nice good explanation 👍

PR #2370, opened 2026-04-21
[#2318] remove all remnants of dark & light themes

agorer:
Closes #2318


groninge01 (approved):

groninge01:
looks like `next-themes` can be removed completely now?

agorer:
> looks like `next-themes` can be removed completely now?

it is still used to get the info from the `themes/base` folder

groninge01:
> it is still used to get the info from the `themes/base` folder

where is that?

groninge01:
also update to just 1 icon here and remove the rest?

https://github.com/balancer/frontend-monorepo/blob/3cd3806a2494f84968813e915655970773943975/apps/frontend-v3/app/layout.tsx#L19-L33

agorer:
> > it is still used to get the info from the `themes/base` folder
> 
> where is that?

`packages/lib/shared/services/chakra/themes/base/*`

agorer:
> also update to just 1 icon here and remove the rest?

This also have to do with the theme of the OS UI, not only about the Balancer app.

PR #2375, opened 2026-04-21
Remove NEXT_PRIVATE_ALCHEMY_KEY env var

agorer:
After merging it can be removed from Github secrets.


groninge01:
and vercel...

groninge01 (approved):

PR #2383, opened 2026-04-23
docs: expand AGENTS.md

MattPereira:
## Summary
- Expands `AGENTS.md` from a single Next.js rule into full onboarding content for coding agents (Claude Code, Cursor, Copilot, etc.)
- Adds build/dev commands, testing workflows (including single-file invocation), and a monorepo architecture overview
- Keeps the Next.js "read local docs before coding" rule prominent near the top

## Why this helps
- **Faster ramp per session** — agents pick up conventions on first read instead of re-discovering commands, folder layout, and patterns every time
- **Shared source of truth** — lives in-repo and applies to every agent tool, rather than being hidden in per-developer config files
- **Surfaces non-obvious details** — single-file integration test invocation, pool-action subdir shape (`form/handlers/queries/modal/`), GraphQL codegen pipeline, and where generated types land — things that otherwise require reading several files to figure out
- **Reduces command-lookup friction** — fork targets, filtered test runs, and the `lint` vs `lint:all` distinction are written down once

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,79 +14,35 @@ Both apps share `packages/lib`; the active project is resolved from `NEXT_PUBLIC
 
 - Use `PROJECT_CONFIG.projectName`, `projectUrl`, `projectLogo` instead of literal `"Balancer"` / `"Beets"` / domain strings.
 - Gate project-only features (maBEETS, relics, etc.) with `isBalancer` / `isBeets`.
-- New config fields go in `config.types.ts` and must be populated in **both** `projects/balancer.ts` and `projects/beets.ts`.lines
+- New config fields go in `config.types.ts` and must be populated in **both** `projects/balancer.ts` and `projects/beets.ts`.
 
 ## Architecture
 
-**Monorepo** using pnpm workspaces + Turborepo.
-
-### Apps (Next.js with Turbopack, React, Chakra UI)
-
-- `apps/frontend-v3` — Balancer app (balancer.fi). Next.js App Router in `app/`.
-- `apps/beets-frontend-v3` — Beets app (same stack, different branding/config).
-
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -1,7 +1,92 @@
...
 
 Before any Next.js work, find and read the relevant doc in `node_modules/next/dist/docs/`. Your training data is outdated — the docs are the source of truth.
 
-<!-- END:nextjs-agent-rules -->
+### Never hardcode project-specific values in `packages/lib`
+
+Both apps share `packages/lib`; the active project is resolved from `NEXT_PUBLIC_PROJECT_ID` in `config/getProjectConfig.ts`, which exposes `PROJECT_CONFIG` and `isBalancer` / `isBeets`. Hardcoding breaks the other app silently.
+
+- Use `PROJECT_CONFIG.projectName`, `projectUrl`, `projectLogo` instead of literal `"Balancer"` / `"Beets"` / domain strings.
+- Gate project-only features (veBAL, LBP, maBEETS, etc.) with `isBalancer` / `isBeets`.
```
groninge01: `LBP` is possibly shared

groninge01 (approved):

AGENTS.md
```diff
@@ -1,7 +1,92 @@
...
+# AGENTS.md
 
-# Next.js: ALWAYS read docs before coding
+This file provides guidance to coding agents working in this repository.
+
+## Rules
+
+### Next.js: ALWAYS read docs before coding
 
 Before any Next.js work, find and read the relevant doc in `node_modules/next/dist/docs/`. Your training data is outdated — the docs are the source of truth.
```
groninge01: I noticed the folder is not found because the AI will search from the same folder as the `AGENTS.md` file is located. 
So change the path here to `apps/frontend-v3/node_modules/next/dist/docs/` and then it should find the docs.

MattPereira: good call tested w/ claude to confirm the next docs integration :tada:

Issue #2395, opened 2026-04-28
Check for bignumber.js usage with STRICT set to true


groninge01:
Add STRICT configuration option: if true (default), throw an exception on invalid input. if false, return NaN on invalid input. ([source](https://github.com/MikeMcl/bignumber.js/blob/main/CHANGELOG.md#1100))

Check all bignumber callers if they can handle a throw instead of NaN (which is the current behaviour).

PR #2396, opened 2026-04-28
chore: upgrade bignumber js

groninge01:
- bump bignumber.js from 9.3.1 to 11.0.0

bignumber.js is more strict so:
- need to import type where needed (no longer added to global scope)
- invalid input behaviour changed for default, manually override to v9 behaviour for now (see also [here](https://github.com/balancer/frontend-monorepo/issues/2395))


groninge01:
did a few adds/removes/swaps, all were successfull

agorer (approved):

PR #2402, opened 2026-04-30
fix: load env on localhost for integration test

groninge01:
on localhost:

- copy `.env.template` in root to `.env.local` (also in root)
- run `pnpm run test:integration` from root

and also:
- added some timeouts & retries fo ci (is ubuntu slower than macos?)
- fixed a codeql alert

### DIFF
--- a/packages/test/anvil/anvil-global-setup.ts
+++ b/packages/test/anvil/anvil-global-setup.ts
@@ -7,6 +7,8 @@ import { ANVIL_NETWORKS, getForkUrl } from './anvil-setup'
 import { testChains } from './testWagmiConfig'
 
 const currentDir = dirname(fileURLToPath(import.meta.url))
+const DEFAULT_ANVIL_START_TIMEOUT_MS = 60_000
+const DEFAULT_ANVIL_STOP_TIMEOUT_MS = 20_000
 
 function loadEnvFileIfExists(path: string) {
   try {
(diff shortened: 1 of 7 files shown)


packages/test/anvil/anvil-global-setup.ts
```diff
@@ -1,8 +1,25 @@
...
+
+function loadEnvFileIfExists(path: string) {
+  try {
+    loadEnvFile(path)
+  } catch (error) {
+    if ((error as NodeJS.ErrnoException).code !== 'ENOENT') throw error
+  }
+}
+
+loadEnvFileIfExists(resolve(currentDir, '../../../.env.local'))
```
agorer, line 21: would this overwrote `.env.local` if the dev already has one with its own configurations?

packages/test/anvil/anvil-global-setup.ts
```diff
@@ -16,12 +33,15 @@ export async function setup() {
...
       forkUrl: getForkUrl(chain.id, false),
       forkBlockNumber: ANVIL_NETWORKS[chain.id].forkBlockNumber,
+      startTimeout: DEFAULT_ANVIL_START_TIMEOUT_MS,
     })
     promises.push(
       startProxy({
         port: chain.port,
         host: '::',
         options: {
+          startTimeout: DEFAULT_ANVIL_START_TIMEOUT_MS,
```
agorer, line 43: Never had problems with timeouts locally.

agorer (approved):

packages/test/anvil/anvil-global-setup.ts
groninge01: no, it only reads the file

packages/test/anvil/anvil-global-setup.ts
groninge01: the timeouts are on ci

MattPereira:
> added some timeouts & retries fo ci (is ubuntu slower than macos?)

i made the argument to switch because i noticed ubuntu was faster #2382

my hypothesis was everyone trying to use mac-os so the queue to run is way longer

i guess you might try switching back and monitoring for yourself

PR #2439, opened 2026-05-13
feat: enable bignumber.js STRICT mode with safe fallback

### LINKED ISSUE #2395: Check for bignumber.js usage with STRICT set to true

groninge01:
## Summary

Enables bignumber.js `STRICT` mode globally while hardening the central `bn()` helper to preserve backwards compatibility for all callers.

Closes #2395

## Problem

bignumber.js v11.0.0 introduced a `STRICT` configuration option (default `true`) that changes behavior when invalid input is passed to the constructor:
- `STRICT: true` → **throws** on invalid input
- `STRICT: false` → returns `NaN` on invalid input

The codebase was explicitly disabling `STRICT` with a TODO to audit all callers before enabling it.

### 1. Enable STRICT globally

Replaces `BigNumber.set({ STRICT: false })` with `BigNumber.set({ STRICT: true })`. This means any **direct** use of `new BigNumber(...)` will now throw on invalid input, catching bugs early.

### 2. Harden `bn()` helper with safe fallback

The `bn()` function in `packages/lib/shared/utils/numbers.ts` is the single entry point for **~338+ calls** across 101 files. It now:
- **Throws** on `null` / `undefined` (programmer errors should fail fast)
- **Catches** invalid string inputs (e.g. `''`, `' '`, `'abc'`) and returns `NaN` explicitly, preserving the existing behavior that callers rely on

```ts
export function bn(val: Numberish): BigNumber {
  if (val == null) {
    throw new TypeError(`Cannot create BigNumber from ${val}`)
  }
  try {
    return new BigNumber(val.toString())
  } catch {
    // Preserve backwards compatibility: callers expect NaN for invalid input
    return new BigNumber(NaN)
  }
}
```

### 3. Update tests

- Renamed test to reflect the new implementation
- Added test verifying `STRICT: true` is active globally

## Why not Option B (fix all call sites)?

Auditing and fixing all ~338+ `bn()` call sites would require touching 50+ files with high regression risk. The current approach is a ~10-line change in a single file that:
- Resolves the issue
- Maintains backward compatibility
- Can be verified in a single PR
- Allows gradual migration of individual call sites in future PRs if desired

## Verification

- [x] All 402 unit tests pass (68 test files)
- [x] TypeScript type-check passes
- [x] ESLint passes
- [x] Prettier formatting passes

### DIFF
--- a/packages/lib/shared/utils/numbers.spec.ts
+++ b/packages/lib/shared/utils/numbers.spec.ts
@@ -1,3 +1,4 @@
+import BigNumber from 'bignumber.js'
 import {
   bn,
   fNum,
@@ -163,7 +164,7 @@ describe('bn', () => {
     expect(bn('0.0000000000000035').toFixed()).toBe('0.0000000000000035')
   })
 
-  test('returns NaN for invalid string inputs instead of throwing', () => {
+  test('returns NaN for invalid string inputs', () => {
     expect(bn('').isNaN()).toBe(true)
     expect(bn(' ').isNaN()).toBe(true)
     expect(bn('abc').isNaN()).toBe(true)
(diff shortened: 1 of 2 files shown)


groninge01:
Closing to recreate with correctly named branch

PR #2536, opened 2026-06-22
chore: bump b-sdk

groninge01:


groninge01:
tested on preview with `Use Signatures` set to `on` -> works!

PR #2547, opened 2026-07-03
feat: Balancer Analytics

Xeonus:
## Feature Set / Description
Adds a dedicated analytics app to apps/balancer-analytics

- standalone, read-only analytics surface for Balancer v2 & v3
- aggregates protocol-wide metrics, per-pool analytics, portfolio views, and veBAL governance data, served from the Balancer API v3, a Neon-backed indexer, and on-chain reads
- Impact on existing packages is minimal and additive

Features:

1. protocol overview page
2. pool page
3. portfolio page
4. governance page

## IMPORTANT: Vercel migration plan & Domain Setup
Once this app is accepted, I propose that a fresh instance is set up for the Neon database, then I can backfill it as there is no straight-forward way to do that apparently. The rest can be migrated over from my vercel org (which currently is hosted via defilytica)

I also propose to host this app at analytics.balancer.fi
Once that is set up, we should do another PR to remove the old DeFilytica references in the main app

## Integration tests
No integration tests yet implemented


groninge01:
new PR is #2549

PR #2560, opened 2026-07-09
chore: remove zkevm chain references from analytics app

groninge01:


Xeonus (approved):
LGTM, no impact on db and other dependencies. Thx for the clean-up

PR #2567, opened 2026-07-17
feat: BPT price history and data fetching improvements

Xeonus:
**Feature addition**                                                                                                                                                                                                                              
  As per Tiago's request, add a BPT pricing history chart                                                                                                                                                                                           
                                                                                                                                                                                                                                                    
  The HODL basket fixes per-BPT token quantities at the start of the window and values them forward. For boosted pools it converts to the **underlying** token — valuing at the wrapped token would credit the yield to both LP and HODL and cancel 
  it out, which defeats the point of the comparison.                                                                                                                                                                                                
                                                                                                                                                                                                                                                    
  **General improvements**                                                                                                                                                                                                                          
  Given the BPT charting introduces another query, I had to clean up how data is fetched in the analytics app. These changes include:                                                                                                               
                                                                                                                                                                                                                                                    
  - **Pool picker modal was hitting api-v3 on every pool page visit.** It was mounted unconditionally, so its 500-row `GetPools` fired straight from the browser even though most people never open it. Now only mounted when open.                 
  - **Landing page fanned out 12 direct api-v3 calls** — one `protocolMetricsChain` per chain, from the browser, uncached. Replaced with a new `/api/protocol-stats-by-chain` route that aliases every chain into a single upstream query and caches
  it. 12 uncached POSTs → 1 cached GET → 1 upstream request.                                                                                                                                                                                        
  - **Pool page state reads had no cache at all.** The page is `force-dynamic`, so the multicalls re-ran on every request — including every range toggle, even though nothing in them depends on the range. Now behind `unstable_cache` keyed on    
  pool identity, so switching 30d/90d/1y costs zero RPC. `?refresh` still bypasses it.                                                                                                                                                              
  - **Enabled multicall batching on the drpc client.** Each state read opened its own HTTP request; a boosted stable pool did 4 round trips per render, now 1. No call-site changes.                                                                
  - **`/api/pool/.../order-flow` was never actually cached.** Only the `s-maxage` header protected it, so every CDN miss (and every dev request) re-paginated ~20 api-v3 calls + ~100 drpc lookups for a 30-day window that's 29/30 immutable.      
  Wrapped in `unstable_cache`.                                                                                                                                                                                                                      
  - **Deduped hooks that fire the same request twice.** Portfolio mounts merkl + gauge-rewards from two separate cards (4 requests for 2 payloads), governance does the same with bal-supply. Added a small module-level request cache              
  (`lib/upstream/request-cache.ts`), generalized from the one `useProtocolSnapshots` already had inline.                                                                                                                                            
  - **Bumped `/api/token-prices` TTL from 5min to 1h.** The series is bucketed per UTC day, so every bucket but today's is frozen — the short TTL was re-pulling a full year of prices to move one trailing point.                                  
  - **Fixed the misleading rate-limit warning.** Browsing a lot of pools trips api-v3's limit, but the notice claimed the upstream was returning errors and that it was "not a problem with your connection or with the pool" — which is wrong and  
  unhelpful. It now names throttling as the likely cause.                                                                                                                                                                                           
                                                                                                                                                                                                                                                    
  Minor: the pool title was clipping descenders (the `special` heading variant paints via `background-clip: text`), added 5px.


tiago-hansen:
New feature looks good 👍

groninge01:
empryo w/ glm-5.2 says

**1. `useProtocolSnapshots.ts` not migrated to `request-cache.ts` (oversight)**
`request-cache.ts:6` says it "Generalizes the store that `lib/snapshots/useProtocolSnapshots.ts` grew inline" — but `useProtocolSnapshots` still has its own inline `loadSnapshots`/`readCached`/`cache` Map (lines 36–82). The generalization was created but the original wasn't migrated. Refactor it to use `dedupedLoad`/`peekCached` for consistency — otherwise the next dev sees two implementations of the same pattern in the same app.

**2. `useMerklRewards` / `useGaugeRewards` missing `peekCached` seeding (UX inconsistency)**
`useBalSupplyByChain` and `useChainProtocolStats` seed `useState` from `peekCached` so a remount with fresh module-cache data shows instantly (no loading flicker). `useMerklRewards` and `useGaugeRewards` don't — they start with `asyncState = null` and return `loading: true` for one render even when `dedupedLoad` resolves synchronously from cache. Add the same `peekCached` init pattern to both.

**3. No tests for `computeHodl` (biggest gap)**
`computeHodl` (PoolBptPriceHistory.tsx:269–324) is the PR's core business logic — pure function, no React, no I/O. It handles: t₀ basket fixing, wrapped→underlying conversion, null-gap walking, and base-value anchoring. It's exactly the kind of function that silently breaks when someone tweaks the math. At minimum, snapshot tests for: plain pool, boosted pool (ERC4626), token with missing price series, pool younger than range. 1691 lines of new code with zero tests.

plus some nits...

Xeonus:
@groninge01 thx for the review, addressed comments [here](https://github.com/balancer/frontend-monorepo/pull/2567/changes/5847bf1a79530645c79d90584e799ebf18769353)

groninge01:
lgtm

fyi: renamed the test file and added the npm scripts to run it in ci

image

groninge01 (approved):

groninge01:
the rest of the analytics app could also use some more tests ;)

PR #2569, opened 2026-07-22
fix(analytics): stop hiding pool table columns, scroll horizontally

gosuto-inzasheru:
we are currently losing some pool columns at certain widths; this fixes that by using making the table scrollable (same as mobile)


Xeonus (approved):
Nice fix, consistent with main app now!

PR #2624, opened 2026-08-06
chore: deprecate mode and fraxtal

groninge01:
Mirrors the zkevm deprecation (#2552) for Mode and Fraxtal chains.

Removes Mode and Fraxtal from the runtime GqlChainValues enum, network configs, RPC maps, wagmi chain map, drpc client, ecosystem pool activity gradients, token colors, CSP issue list, pool creation network filter, and the analytics app's chain maps. Deletes the network config files and chain SVG icons.

isChainDeprecated() now returns false (no chains are deprecated). The GraphQL schema still defines MODE and FRAXTAL — the spec allowlist in graphql-enums.spec.ts now includes them alongside ZKEVM so the runtime-vs-schema check passes.

Closes #2235

PR #2635, opened 2026-08-07
chore: remove XLayer chain support

### LINKED ISSUE #1750: Add support for X-Layer (OKX) chain
- Homepage: https://web3.okx.com/xlayer
- Explorer: https://www.oklink.com/x-layer, https://web3.okx.com/explorer/x-layer
- Viem conf: https://github.com/wevm/viem/blob/main/src/chains/definitions/xLayer.ts
- Sample PR: https://github.com/balancer/frontend-monorepo/pull/1591/files

groninge01:
Closes #1750

Removes XLayer from the runtime GqlChainValues enum, network configs, RPC maps, wagmi chain map, drpc client names, CSP issue list, token colors, pool creation networks, fork test balances, and the analytics app's chain maps. Deletes the network config file, chain SVG icons, and the Makefile fork target.

The GraphQL schema still defines XLAYER — the spec allowlist in graphql-enums.spec.ts now includes it alongside ZKEVM so the runtime-vs-schema check passes.

Mirrors the Mode/Fraxtal deprecation (#2624).


groninge01:
@Xeonus fyi: also removing x-layer support (that was never on prod)

PR #2711, opened 2026-08-26
test(analytics): add unit tests to balancer-analytics

groninge01:
## What

- Added 13 unit-test specs to `apps/balancer-analytics`, covering the app's pure-logic modules: dune normalization, CoW fee-rebate stripping, pool param snapshots, event-log decoding, fetch retry, request cache, portfolio P&L, veBAL holder snapshot, Dune label pagination, formatters, block-window helpers, and drpc log chunking / block timestamps.
- Fixed the analytics vitest config so the suite runs standalone: set `NEXT_PUBLIC_BALANCER_API_URL` in `test.env` (was throwing at import via `@repo/lib/config/app.config.ts`), added the `@analytics` path alias, and aliased Next's `server-only` virtual module to a no-op stub.
- Exported `computePnl` in `lib/hooks/usePortfolioPnl.ts` (previously internal) so its P&L/IL logic is unit-testable.

## Why

- Issue #2710 requested meaningful tests for `balancer-analytics`. The app had only 2 spec files, and both failed to collect because the vitest config didn't provide the env vars `@repo/lib` requires at import. The suite now runs green with 17 files / 110 tests.

## Test Steps

- [ ] `pnpm --filter balancer-analytics test:unit` — all 17 files / 110 tests pass
- [ ] `pnpm --filter balancer-analytics typecheck` — clean
- [ ] `pnpm --filter balancer-analytics lint` — clean

## Risks / Breaking Changes

- `computePnl` is now exported from `usePortfolioPnl.ts` — additive, no behavior change.
- Vitest config changes are test-only; no runtime impact on the app.
- Does not touch `packages/lib`, so no impact on the Balancer/Beets apps.

## Other Notes

- Integration tests (sync orchestrator, API routes) and additional e2e smoke tests were deliberately deferred — the app is read-only and the repo's anvil-based integration harness is a poor fit; the pure logic underneath is already unit-tested.

### DIFF
--- a/apps/balancer-analytics/app/_components/format.spec.ts
+++ b/apps/balancer-analytics/app/_components/format.spec.ts
@@ -1,0 +1,16 @@
+import { describe, expect, it } from 'vitest'
+import { usd } from './format'
+
+describe('usd', () => {
+  it('formats large values compactly', () => {
+    expect(usd(5_400_000_000)).toBe('$5.4B')
+  })
+
+  it('formats zero as $0', () => {
+    expect(usd(0)).toBe('$0')
+  })
+
+  it('formats negative values', () => {
+    expect(usd(-1_200_000)).toBe('-$1.2M')
+  })
+})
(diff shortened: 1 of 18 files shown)


gosuto-inzasheru:
@cursoragent check out these new tests, vitest config, pnl computation and any other additions that this pr introduces!

gosuto-inzasheru (approved):
lgtc

PR #2736, opened 2026-09-01
Fix LBP shared-link start time timezone hydration

tiago-hansen:
## What

- Added `unixTimestampToDateTimeLocalString` in `packages/lib/shared/utils/time.ts` to format on-chain unix timestamps for `datetime-local` inputs in the user's local timezone
- Updated `useHydrateLbpForm` to use the new helper instead of slicing a UTC ISO string when loading a deployed LBP from a shared initialize link (both the dynamic and fixed-price paths)
- Updated the LBP e2e helpers (`create-lbp.helpers.ts`, `create-lbp.spec.ts`) to fill `datetime-local` inputs through the same helper, replacing the `toISOString(...).slice(0, 16)` pattern
- Refactored the helper to build its date with `fromUnixTime` instead of a manual seconds-to-milliseconds conversion
- Removed the now-unused `toISOString` export from `packages/lib/shared/utils/time.ts`
- Added unit tests covering local-time formatting and round-trip conversion with `dateTimeToUnixTimestampBigInt`

## Why

When a deployed LBP is opened via a shared link (e.g. `/lbp/create/MAINNET/0x…`), start/end times are hydrated from on-chain UTC timestamps. The previous implementation used `toISOString(...).slice(0, 16)`, which produced a string like `2026-09-01T18:00` without timezone information. Browsers and `parseISO` interpret that as local wall time, so users in CET saw 18:00 UTC as 18:00 local (16:00 UTC) and hidden start-time validation failed silently — blocking **Next** even when the sale was still in the future.

The e2e helpers fed the same `datetime-local` inputs with the same UTC-sliced string, so the LBP creation tests only passed because the Playwright container runs in UTC; on a machine with a non-UTC timezone the filled start time lands in the past and step 1 rejects it. Routing both the app and the tests through one helper keeps the two in agreement.

## Test Steps

- [ ] Deploy an LBP without seeding and copy the initialize link
- [ ] Open the link in a browser set to a non-UTC timezone (e.g. Europe/Berlin)
- [ ] Connect a wallet with enough launch tokens, enter the sale amount, and click **Next**
- [ ] Confirm step 2 loads instead of the page staying on step 1 with no visible error
- [ ] Confirm the start/end time fields show the sale times in your local timezone, not UTC
- [ ] Run `pnpm --filter @repo/lib exec vitest run shared/utils/time.spec.ts` and confirm all tests pass
- [ ] Confirm `Unit-Test` and `E2E-Dev-Test-Balancer-2` (which runs `create-lbp.spec.ts`) pass on this PR

## Risks / Breaking Changes

- Touches shared `packages/lib` code used by both Balancer and Beets apps (`NEXT_PUBLIC_PROJECT_ID`), but only affects LBP form hydration from shared initialize links. Beets mounts the same `LbpCreationPage` at `/lbp/create` (its nav entry is gated to dev/staging in `packages/lib/config/projects/beets.ts:51`), so the fix applies there too rather than being Balancer-only.
- No change to how users pick dates when creating a new LBP from scratch; creation and initialize flows still round-trip through `dateTimeToUnixTimestampBigInt`.
- `toISOString` is removed from `packages/lib/shared/utils/time.ts`. It had no remaining callers in the repo; code needing a raw ISO string should use `Date.prototype.toISOString` directly.
- E2E coverage for the helper changes is only exercised against a dev server plus an anvil fork, so it is verified by CI rather than locally.

## Other Notes

`packages/lib/shared/utils/time.spec.ts` asserts an exact local-time string, which is only stable because `packages/lib/test/vitest/setup-vitest.tsx` pins `process.env.TZ` to `Europe/Madrid`. The comment there names that dependency; the round-trip test alongside it is timezone-agnostic.

<div><a href="https://cursor.com/agents/bc-d7e934cc-9d33-41f3-90a9-949b9d460764?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d7e934cc-9d33-41f3-90a9-949b9d460764&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/packages/e2e-tests/helpers/create-lbp.helpers.ts
+++ b/packages/e2e-tests/helpers/create-lbp.helpers.ts
@@ -1,7 +1,12 @@
 import { clickButton, button, clickRadio, checkbox } from '@/helpers/user.helpers'
 import { expect, type Page } from '@playwright/test'
 import { LBP_FORM_STEPS } from '@repo/lib/modules/lbp/constants.lbp'
-import { oneDayInMs, oneWeekInMs, toISOString } from '@repo/lib/shared/utils/time'
+import {
+  oneDayInMs,
+  oneWeekInMs,
+  unixTimestampToDateTimeLocalString,
+  oneSecondInMs,
+} from '@repo/lib/shared/utils/time'
 
 export type LbpSaleType = 'seedless' | 'seeded' | 'fixed-price'
 export type LbpConfig = {
@@ -91,8 +96,12 @@ export async function doSaleStructureStep(
 
   await expect(page.getByRole('heading', { name: 'Sale period' })).toBeVisible()
   const dateInputs = page.locator('input[type="datetime-local"]')
-  await dateInputs.first().fill(toISOString(Date.now() + oneDayInMs).slice(0, 16))
-  await dateInputs.last().fill(toISOString(Date.now() + oneWeekInMs).slice(0, 16))
+  await dateInputs
+    .first()
+    .fill(unixTimestampToDateTimeLocalString((Date.now() + oneDayInMs) / oneSecondInMs))
+  await dateInputs
+    .last()
+    .fill(unixTimestampToDateTimeLocalString((Date.now() + oneWeekInMs) / oneSecondInMs))
 
   if (lbpConfig.saleType === 'seedless') {
     await expect(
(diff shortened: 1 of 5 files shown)


groninge01:
added a few small refactors -> lgtm

groninge01 (approved):

PR #2774, opened 2026-09-09
chore(analytics): remove obsolete graphql codegen and dead abis

groninge01:
## What
- Removed the app-local GraphQL codegen: `codegen.ts`, `lib/generated/`, and `lib/services/subgraph/` (ormilabs v3-vault subgraph client, endpoints, and the `GetRecentActivity` query) along with the `graphql:gen` script and its `@graphql-codegen/cli`, `@graphql-codegen/client-preset`, and `graphql` devDependencies.
- Removed the `lib/abis/` directory — 7 full deployed ABIs (v2/v3 vault, fee controller, stable pool, surge hook, weighted pool, ~5.7k LOC) that nothing imported.
- Moved the one used export, `V3_VAULT_ADDRESS`, from `lib/abis/v3-vault.ts` into `lib/contracts/v3-addresses.ts` and re-pointed `lib/pool-events/sync.ts` at it.
- Cleaned up the config surfaces that referenced the deleted paths: eslint ignores, tsconfig exclude, `.prettierignore`, and the build/dev scripts in `package.json` (build/dev now run only the shared `@repo/lib` codegen).
- Updated stale doc comments in `lib/pool-events/event-signatures.ts` and the app README.

## Why
The balancer-analytics app was migrated onto api-v3 (`@repo/lib`) and raw RPC for all data paths, leaving two dead subsystems behind:
- The v3-vault subgraph codegen's only query (`GetRecentActivity`) had zero consumers anywhere in the app.
- The full ABI files were never imported — live code uses inline `parseAbi` getters (`lib/pool-state/read.ts`) and the parseAbi event tuples in `lib/pool-events/event-signatures.ts`.

Net: ~5,900 lines deleted.

## Test Steps
No user-facing surface changes — this is pure dead-code removal. Verification performed:
- [x] `pnpm --filter balancer-analytics typecheck` passes
- [ ] `pnpm --filter balancer-analytics lint` passes
- [x] `pnpm --filter balancer-analytics test:unit` passes (110 tests)
- [x] `pnpm --filter balancer-analytics build` completes (routes render; dev/build no longer run the app-local codegen)

## Risks / Breaking Changes
- `build`/`dev` previously ran the app-local codegen before starting; they now rely solely on `@repo/lib`'s generated api-v3 types, which were already generated in both flows. If the ormilabs v3-vault subgraph is ever needed again, the codegen config and its graphql documents would have to be reintroduced.
- Does not touch `packages/lib` — no cross-app (Balancer/Beets) impact.

PR #2775, opened 2026-09-10
fix: analytics cleanup followups

gosuto-inzasheru:
@groninge01 see https://github.com/balancer/frontend-monorepo/pull/2774 for context


groninge01:
#2774 is updated

